PR #7, opened 2020-12-13
feat: Handle account code in generator

TheWaWaR:
* Add 2 syscalls:
  - `SYS_LOAD_DATA`
  - `SYS_STORE_DATA`
* Add 3 functions to gw_context_t
  - `sys_load_nonce`
  - `sys_load_data`
  - `sys_store_data`
*

### DIFF
--- a/c/common.h
+++ b/c/common.h
@@ -61,10 +61,20 @@ void gw_build_account_key(uint32_t id, const uint8_t key[GW_KEY_BYTES],
   blake2b_final(&blake2b_ctx, raw_key, GW_KEY_BYTES);
 }
 
+void gw_build_account_field_key(uint32_t id,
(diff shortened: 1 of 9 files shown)


jjyr:
Our need is to store the EVM bytecode into Godwoken, but we can provide a more generalized feature: to store/load any data into Godwoken.

In this PR, we implement a map `codes[account_id] -> code`, to change it to a more generalized version we can implement the map as `data[key] -> data` which the key is `ckb_blake2b(data)`. And rename the two syscalls `SYS_STORE/LOAD_DATA`.

Benefits:

* Provided the same feature
* An account can load any data, not only the code

TheWaWaR:
Changed follow the suggestions.

jjyr (approved):

PR #9, opened 2020-12-17
feat: godwoken js binding

bitrocks:


crates/setup-tool/src/main.rs
```diff
@@ -136,7 +136,6 @@ fn run() -> Result<()> {
     let rpc_listen_address = args.value_of(RPC_LISTEN_ADDRESS).unwrap();
     let lumos_callback_address = args.value_of(LUMOS_CALLBACK_ADDRESS).unwrap();
     let lumos_endpoint = args.value_of(LUMOS_ENDPOINT).unwrap();
-    let rollup_contract_path = args.value_of(ROLLUP_CONTRACT).unwrap();
```
jjyr, line 139: Why?

bitrocks: Oops, I added some new config options below here and considered unnecessary so far. While  I remove those codes and delete this line by accident. Gonna fix it.

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,125 @@
...
+export type SyncEvent =
+  | "Success"
+  | "BadBlock"
+  | "BadChallenge"
+  | "WaitChallenge";
+
+export type Status = "Running" | "Halting";
+
+export class ChainService {
+  sync(syncParam: SyncParam): SyncEvent;
```
xxuejie, line 120: How do I passa L2 transaction to the chain service?

bitrocks: Gonna add a `send_transaction` here soon.

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export interface L1Action {
+  transaction_info: TransactionInfo;
+  header_info: HeaderInfo;
+  context: SubmitTxs | StartChallenge | CancelChallenge | Revert;
+}
+
+export interface NextBlockContext {
+  aggregator_id: HexNumber;
+  timestammp: HexNumber;
```
xxuejie, line 16: `timestamp` is mis-spelled

packages/godwoken/native/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    pub fn stop(&self) {
+        self.running.store(false, Ordering::Release);
+    }
+}
+
+declare_types! {
+    pub class JsNativeChain for NativeChain {
+        init(mut cx) {
+            let config_path = cx.argument::<JsString>(0)?.value();
```
xxuejie, line 52: I think genesis configuration can also be passed from JS side to Rust side. This way we can manage all configurations in JS, for a more consistent design.

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,141 @@
...
+export interface SyncParam {
+  reverts: L1Action[];
+  updates: L1Action[];
+  next_block_context: NextBlockContext;
+}
+
+export interface L1Action {
+  transaction_info: TransactionInfo;
+  header_info: HeaderInfo;
+  context: SubmitTxs | StartChallenge | CancelChallenge | Revert;
```
xxuejie, line 11: Using [discriminating unions](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) here would be better.

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,141 @@
...
+  args: ArrayBuffer;
+}
+
+export class ChainService {
+  constructor();
+  sync(syncParam: SyncParam): Promise<SyncEvent>;
+  produce_block(
+    produceBlockParam: ProduceBlockParam
+  ): Promise<L2BlockWithState>;
+  submitL2Transaction(l2Transaction: L2Transaction): Promise<void>;
```
xxuejie, line 137: Maybe we can have `RunResult` returned here: https://github.com/nervosnetwork/godwoken/blob/b28d7ccfcd62a4ac26e3a1b0646866657d7ca958/crates/generator/src/generator.rs#L134-L139?

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+export interface RawL2Transaction {
+  from_id: HexNumber;
+  to_id: HexNumber;
+  nonce: HexNumber;
+  args: ArrayBuffer;
+}
+
+export class ChainService {
```
xxuejie, line 131: We also need a public method to get current storage state, much like Ethereum's `eth_getStorageAt`

packages/godwoken/index.d.ts
bitrocks: fixed.

packages/godwoken/native/src/lib.rs
bitrocks: Have updated the TS definition to support configuration.

packages/godwoken/index.d.ts
bitrocks: Updated.

packages/godwoken/index.d.ts
bitrocks: Gonna add a separate `execute` method return `RunResult` later.

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,181 @@
...
+  timestamp: HexNumber;
+}
+
+export interface AggregatorConfig {
+  account_id: HexNumber;
+  signer: SignerConfig;
+}
+
+export interface SignerConfig {}
+
```
xxuejie, line 170: One additional feature: we need to expose the method to build genesis block, so we can attach code from JS side for a genesis creation tool

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,181 @@
...
+  header_info: HeaderInfo;
+  context: SubmitTxs | StartChallenge | CancelChallenge | Revert;
+}
+
+export interface NextBlockContext {
+  aggregator_id: HexNumber;
+  timestamp: HexNumber;
+}
+
+export interface TransactionInfo {
```
xxuejie, line 19: It would be better if the JS API can only require Transaction type. We can do the conversion from Transaction to TransactionInfo internally at Rust side.

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,181 @@
...
+  transaction: Transaction;
+  block_hash: Hash;
+}
+
+export interface HeaderInfo {
```
xxuejie, line 24: Same as the above, it would be better if the JS API only requires Header, instead of HeaderInfo

crates/common/src/lib.rs
```diff
@@ -17,7 +17,10 @@ pub use sparse_merkle_tree;
 
 pub const DEPOSITION_CODE_HASH: [u8; 32] = [0u8; 32];
 pub const SUDT_CODE_HASH: [u8; 32] = [0u8; 32];
-pub const CKB_SUDT_SCRIPT_HASH: [u8; 32] = [0u8; 32];
+pub const CKB_SUDT_SCRIPT_HASH: [u8; 32] = [
```
xxuejie, line 20: Where did this come from?

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export class ChainService {
+  constructor(config: Config);
+  sync(syncParam: SyncParam): Promise<SyncEvent>;
+  produce_block(
+    produceBlockParam: ProduceBlockParam
+  ): Promise<ProduceBlockResult>;
+  submitL2Transaction(l2Transaction: ArrayBuffer): Promise<RunResult>;
+  execute(l2Transaction: ArrayBuffer): Promise<RunResult>;
+  buildGenesisBlock(config: GenesisConfig): Promise<GenesisWithSMTState>;
```
xxuejie, line 144: Building genesis block should be a separate function outside of ChainService: initializing chain service requires config, config contains rollup type hash, but rollup type hash is only available when the genesis block is committed to chain. Doing it this way will result in chicken-egg problem.

packages/godwoken/index.d.ts
```diff
@@ -0,0 +1,150 @@
...
+export interface GenesisWithSMTState {
+  genesis: ArrayBuffer; // gw_types::packed::L2Block
+  // branches_map: Record<Hash, BranchNode>,
+  // leaves_map: Record<Hash, LeafNode<H256>>,
+}
+
+export class ChainService {
+  constructor(config: Config);
+  sync(syncParam: SyncParam): Promise<SyncEvent>;
+  produce_block(
```
xxuejie, line 139: This is missing a package function, which is essentially a wrapper over [this](https://github.com/nervosnetwork/godwoken/blob/2c515f5d13fcec611b1fdbac558b48609e855e64/crates/chain/src/tx_pool.rs#L150), otherwise we won't be able to generate `TxPoolPackage` needed in `ProduceBlockParam`

xxuejie: An alternative path, is to modify `ProduceBlockParam` in JS definition to contain deposition requests instead of `TxPoolPackage`, then internally in this function, we can first invoke package to get `TxPoolPackage`, then call the real produce_block. This way, we won't need to care for `TxPoolPackage` in JS side.

crates/common/src/lib.rs
bitrocks: It's a temp value to pass the `assert` in https://github.com/nervosnetwork/godwoken/blob/2c515f5d13fcec611b1fdbac558b48609e855e64/crates/store/src/genesis.rs#L50-L54

packages/godwoken/index.d.ts
bitrocks: Agreed, it's unnecessary to build `TxPoolPackage`. Gonna update it.

PR #10, opened 2020-12-18
Update generator interface

jjyr:
1. Support executes a tx without push it into the tx pool.
2. Provide a `TxPool.push` like interface for withdrawal requests
    1. Add an interface to verify the withdrawal requests.
    2. Add an interface to push a withdrawal request into the tx pool.
3. `produce_block` function receives TxPoolPackage instead of receiving depositions and withdrawals.
4. Remove withdrawal_requests argument from `Chain.sync` parameters.


xxuejie:
Right now, a chain is initialized by manually passing in `L2Block`, last synced info, as well as state store. This would require an L2Block assembled from the outside. What if we change the API here, so a ChainService is initialized from genesis transaction(of course, when persistent store is built, a ChainService should then be initialized from an existing store)? Personally I feel that is a better abstraction.

xxuejie:
Another question: `TransactionInfo` only occurs together with `HeaderInfo`, in this case, `block_hash` is already provided in `HeaderInfo`, maybe we can just use `Transaction`, not `TransactionInfo`?

jjyr:
Updated @xxuejie

PR #15, opened 2020-12-23
feat: Add withdrawal request

bitrocks:


packages/runner/src/runner.ts
```diff
@@ -32,6 +35,10 @@ import {
 import { AlwaysSuccessGenerator } from "./locks";
 import * as secp256k1 from "secp256k1";
 
+// TODO
```
xxuejie, line 38: Please implement this TODO.

packages/runner/src/runner.ts
```diff
@@ -475,4 +474,346 @@ export class Runner {
...
+        ckbWithdrawalCapacity + BigInt(withdrawalCapacity);
+      const sudtScriptHash = new Reader(
+        rawWithdrawalRequest.getSudtScriptHash().raw()
+      ).serializeJson();
+      let withdrawalType: Script | undefined = undefined;
+      let outputData = "0x";
+      // check if it includes sudt withdrawal request
+      if (
+        sudtScriptHash !=
+        "0x0000000000000000000000000000000000000000000000000000000000000000"
```
xxuejie, line 525: Why do we compare with a all zero script hash? I don't feel we know any script here whose hash is all zeros.

packages/runner/src/runner.ts
```diff
@@ -475,4 +474,346 @@ export class Runner {
...
+        BigInt(outputCkbCapacitySumForSudtCustodianCells)
+    ) {
+      // so lucky with it, just return
+      return txSkeleton;
+    }
+    // build ckb change custodian cell
+    const custodianLock: Script = {
+      code_hash: this._deploymentConfig().custodian_lock.code_hash,
+      hash_type: this._deploymentConfig().custodian_lock.hash_type,
+      args: DEFAULT_CUSTODIAN_LOCK_ARGS,
```
xxuejie, line 672: If this is a change cell, can we just leverage lock args from the original custodian cell?

packages/runner/src/runner.ts
```diff
@@ -475,4 +474,346 @@ export class Runner {
...
+          BigInt(inputCkbCapacitySum) >=
+          BigInt(ckbWithdrawalCapacity) +
+            BigInt(outputCkbCapacitySumForSudtCustodianCells) +
+            BigInt(minimalCapacity)
+        ) {
+          break;
+        }
+      }
+    }
+    // TODO collect ckb from other sudt custodian cells if all non sudt custodian cells' ckb capacity is insufficient,
```
xxuejie, line 719: I think we can add checkings in godwoken: if someone chooses to withdraw SUDT, he/she must withdraw CKB of enough capacity to store the SUDTs. Also, the left CKBs must be enough to hold a left-over custodian cell.

If above rules are not met, we should reject the withdrawal request. In this case, TODO here won't be needed.

packages/runner/src/runner.ts
```diff
@@ -475,4 +474,346 @@ export class Runner {
...
+      BigInt(outputCkbCapacitySumForSudtCustodianCells);
+    ckbChangeCustodian.cell_output.capacity =
+      "0x" + newChangeCapacity.toString(16);
+    txSkeleton = txSkeleton.update("outputs", (outputs) => {
+      return outputs.push(ckbChangeCustodian);
+    });
+    return txSkeleton;
+  }
+
+  _buildWithdrawalLockArgs(
```
xxuejie, line 741: Please write a denormalizer for RawWithdrawalRequest, and a normalizer for WithdrawalLockArgs.

packages/runner/src/runner.ts
bitrocks: In the case where user only withdraws CKB from L2, the `sudt_script_hash` would be simply set as all-zeros, and yes, it's not supposed to be found onchain. (Have confirmed it with @jjyr )

packages/base/src/index.ts
```diff
@@ -6,11 +6,16 @@ import { CellDep, HexString, HexNumber, Hash, Script } from "@ckb-lumos/base";
...
   state_validator_type: Script;
+  sudt_type: Script;
 
   deposition_lock_dep: CellDep;
   custodian_lock_dep: CellDep;
   state_validator_lock_dep: CellDep;
   state_validator_type_dep: CellDep;
+  sudt_type_dep: CellDep;
+
+  l2_finalized_period: BigInt;
```
xxuejie, line 20: This is not a cell configuration, it should be in Godwoken's config, not here.

packages/base/src/index.ts
```diff
@@ -6,11 +6,16 @@ import { CellDep, HexString, HexNumber, Hash, Script } from "@ckb-lumos/base";
 export interface DeploymentConfig {
   deposition_lock: Script;
   custodian_lock: Script;
+  withdrawal_lock: Script;
```
xxuejie, line 9: Add `withdrawal_lock_dep` as well.

packages/base/src/index.ts
xxuejie: https://github.com/nervosnetwork/godwoken/blob/a0d93bc8af5ecadfd9491cdf859803b72e120c0c/crates/common/src/lib.rs#L18 We can set it as a hardcoded value like the Rust side for now, then later refactor it to a config variable.

bitrocks: updated.

packages/base/src/index.ts
bitrocks: updated.

packages/runner/src/runner.ts
bitrocks: updated.

packages/runner/src/runner.ts
bitrocks: After future discussion, we agreed with the default value approach.

packages/runner/src/runner.ts
bitrocks: updated.

packages/runner/src/runner.ts
bitrocks: updated.

crates/common/src/lib.rs
```diff
@@ -27,6 +27,7 @@ pub const CKB_SUDT_SCRIPT_HASH: [u8; 32] = [
 pub const CKB_SUDT_SCRIPT_ARGS: [u8; 32] = [0; 32];
 pub const ACCOUNT_LOCK_CODE_HASH: [u8; 32] = [0u8; 32];
 pub const ROLLUP_LOCK_CODE_HASH: [u8; 32] = [0u8; 32];
+pub const CKB_NON_SUDT_CHECK_ARGS: [u8; 32] = [0u8; 32];
```
jjyr, line 30: `CKB_SUDT_SCRIPT_ARGS` should be enough to determine a SUDT and CKB.

crates/generator/src/generator.rs
```diff
@@ -114,14 +116,26 @@ impl Generator {
...
-        if amount > balance {
+        // check CKB balance
+        let ckb_balance = state.get_sudt_balance(CKB_SUDT_ACCOUNT_ID, id)?;
+        if capacity > ckb_balance as u64 {
             return Err(ValidateError::InvalidWithdrawal.into());
         }
+        // check sudt balance
+        // when sudt_script_hash is [0u8;32] means only withdraw ckb in the withdrawal_request,
+        // otherwise check the sudt existence and balance.
+        if sudt_script_hash != CKB_NON_SUDT_CHECK_ARGS {
```
jjyr, line 127: Use CKB_SUDT_SCRIPT_ARGS

jjyr (changes requested):

crates/generator/src/error.rs
```diff
@@ -62,6 +62,8 @@ pub enum WithdrawalError {
     InvalidNonce { expected: u32, actual: u32 },
     #[error("Withdrawal Faked CKB")]
     WithdrawFakedCKB,
+    #[error("Non positive sudt amount")]
+    NonPositiveSudtAmount,
```
jjyr, line 66: Use `WithdrawalError::Overdraft`.

crates/generator/src/generator.rs
```diff
@@ -129,6 +129,9 @@ impl Generator {
             .ok_or(ValidateError::UnknownSUDT)?;
         if sudt_id != CKB_SUDT_ACCOUNT_ID {
             // check SUDT balance
+            if amount <= 0 {
```
jjyr, line 132: Remove this, the code below already does the verification.

jjyr (changes requested):

crates/chain/src/chain.rs
```diff
@@ -629,6 +629,10 @@ impl Chain {
         };
         // reverted_block_root: Byte32,
         let last_finalized_block_number = number.saturating_sub(FINALIZE_BLOCKS);
+        println!(
```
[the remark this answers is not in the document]
bitrocks: updated.

crates/generator/src/generator.rs
bitrocks: User can't withdrawal 0 SUDT when non-CKB sudt_id exists.

crates/generator/src/error.rs
jjyr, line 66: rename Sudt -> SUDT to keep consistency

bitrocks: updated.

jjyr (approved):

PR #27, opened 2020-12-25
JSONRPC Server implementation

xxuejie:


packages/runner/src/jsonrpc_server.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  constructor(chainService: ChainService, listen: string) {
+    this.chainService = chainService;
+    this.server = new jayson.Server({
+      ckb_submitL2Transaction: this.submitL2Transaction.bind(this),
+      ckb_executeL2Tranaction: this.executeL2Transaction.bind(this),
+      ckb_submitWithdrawalRequest: this.submitWithdrawalRequest.bind(this),
+      ckb_getBalance: this.getBalance.bind(this),
+      ckb_getStorageAt: this.getStorageAt.bind(this),
+      ckb_getAccountIdByScriptHash: this.getAccountIdByScriptHash.bind(this),
```
TheWaWaR, line 28: Why the prefix are all `ckb_` ?

xxuejie: It's to make a clear distinction between them and Ethereum RPCs: https://eth.wiki/json-rpc/API

At later time we might implement Ethereum RPCs in polyjuice for better developer experience. Having prefixes here allows us to tell ETH ones from CKB ones

TheWaWaR: I think it's better use `gw_` prefix, since those are all godwoken related RPC.

xxuejie: Yeah that makes sense, will change them

TheWaWaR (approved):

PR #28, opened 2020-12-26
feat: Add stake-lock contract

bitrocks:


contracts/stake-lock/src/entry.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+pub fn main() -> Result<(), Error> {
+    let (rollup_type_hash, lock_args) = parse_lock_args()?;
+
+    // read global state from rollup cell
+    let global_state = match search_rollup_state(&rollup_type_hash, Source::Dep)? {
+        Some(state) => state,
+        None => return Err(Error::RollupCellNotFound),
```
jjyr, line 52: If a staker tries to update the Rollup cell, then the Rollup cell actually exists in the tx.input, this pattern match will return an error. But according to the verification logic, we should try to unlock by Rollup cell after we failed to unlock by the user.

jjyr (changes requested):

contracts/stake-lock/src/entry.rs
bitrocks: Gonna fix it.

PR #30, opened 2020-12-26
feat: Add ethereum signing module

xxuejie:


crates/generator/src/account_lock_manage/secp256k1.rs
```diff
@@ -58,3 +59,71 @@ impl LockAlgorithm for Secp256k1 {
...
+            return Ok(false);
+        }
+        Ok(true)
+    }
+}
+
+#[test]
+fn test_secp256k1_eth() {
+    let mut hasher = Keccak256::new();
+    hasher.update("\x19Ethereum Signed Message:\n32");
```
jjyr, line 113: Because Rollup passes a unified format message to the different signature algorithms, we should hash the prefix("\x19Ethereum Signed Message:\n | message.len()") in the Secp256k1Eth::verify_signature method.

jjyr (changes requested):

crates/generator/src/account_lock_manage/secp256k1.rs
xxuejie: Agreed, this is now fixed.

crates/generator/src/account_lock_manage/secp256k1.rs
```diff
@@ -58,3 +59,73 @@ impl LockAlgorithm for Secp256k1 {
...
+        &self,
+        lock_args: Bytes,
+        signature: Signature,
+        message: H256,
+    ) -> Result<bool, LockAlgorithmError> {
+        if lock_args.len() != 20 {
+            return Err(LockAlgorithmError::InvalidLockArgs);
+        }
+        let mut hasher = Keccak256::new();
+        hasher.update("\x19Ethereum Signed Message:\n32");
```
jjyr, line 81: hasher.update("\x19Ethereum Signed Message:\n" + message.len());

jjyr (approved):

PR #69, opened 2021-01-08
Support log event

TheWaWaR:


crates/generator/src/types.rs
```diff
@@ -1,10 +1,41 @@
...
+    packed::{self, StartChallenge, StartChallengeWitness},
+    prelude::*,
+};
 use std::{
     collections::HashMap,
     fmt::{self, Display},
 };
 
+#[derive(Debug, PartialEq, Clone, Eq, Default)]
+pub struct LogItem {
```
jjyr, line 13: Can we use `packed::LogItem` directly?

jjyr:
Rebase to master to fix the CI https://github.com/nervosnetwork/godwoken/pull/72

crates/generator/src/types.rs
TheWaWaR: Will be a bit inconsistent with other fields in `RunResult` since they are all native types.

jjyr: Actually, `RunResult` is the only struct that needs to be native because the fields are `HashMap` type. Other structs such as `TxReceipt` are already migrated into the Molecule version.

TheWaWaR: Ok

TheWaWaR:
Fixed and rebased.

jjyr (approved):

PR #77, opened 2021-01-16
Implement get block hash syscall

TheWaWaR:
* Move `trait CodeStore` to gw-store
* Move `Store::init_genesis` to `gw_generator::genesis::init_genesis`
* Implement get block hash syscall


jjyr:
1 Do not move code into gw-common, gw-common is a special crate which used by both scripts and off chain program. This crate should only contains common parts. I d like to move the code into generator or a standalone crate like gw-traits to avoid cyclic dependent.
2. we should add a field store into generator instead of passing it as a argument.

crates/chain/src/chain.rs
```diff
@@ -468,7 +468,11 @@ impl Chain {
         };
         let mut tree = db.account_state_tree()?;
         // process transactions
-        let result = match self.generator.apply_state_transition(&mut tree, args) {
+        let result = match self.generator.apply_state_transition(
+            &self.store.begin_transaction(),
```
jjyr, line 472: We should just pass the `db` to function, to make sure the uncommitted data is readable.

crates/chain/src/mem_pool.rs
```diff
@@ -25,6 +22,7 @@ const MAX_DATA_BYTES_LIMIT: usize = 25_000;
 
 /// TODO remove txs from pool if a new block already contains txs
 pub struct MemPool {
+    store: Store,
```
jjyr, line 25: The mem pool should always use the overlay store. otherwise, the state may be inconsistent.

crates/chain/src/mem_pool.rs
```diff
@@ -108,9 +108,12 @@ impl MemPool {
         self.verify_tx(&tx)?;
         // 2. execute contract
         let raw_tx = tx.raw();
-        let run_result = self
-            .generator
-            .execute(&self.state, &self.next_block_info, &raw_tx)?;
+        let run_result = self.generator.execute(
+            &self.store.begin_transaction(),
```
jjyr, line 112: Just use OverlayStore

jjyr (approved):

jjyr (changes requested):
Please update

jjyr (approved):

PR #78, opened 2021-01-16
Add sys_increase_nonce

TheWaWaR:


jjyr:
What is the usecase of this syscall? why we need to expose increase nonce to contracts?

TheWaWaR:
> What is the usecase of this syscall? why we need to expose increase nonce to contracts?

It's not a syscall. The use case is to increase a contract account's nonce.

PR #80, opened 2021-01-19
feat: Add sentry sdk

bitrocks:


packages/runner/src/index.ts
```diff
@@ -11,6 +11,17 @@ import { Level, RunnerConfig } from "./utils";
...
 import Knex from "knex";
 import deepFreeze from "deep-freeze-strict";
+import * as Sentry from "@sentry/node";
+import * as dotenv from "dotenv";
+import { version } from "../package.json";
+
+dotenv.config();
+
+Sentry.init({
+  dsn: process.env.SENTRY_DSN,
```
jjyr, line 21: We use setting this value from the config file, instead of from env.

bitrocks: updated.

PR #91, opened 2021-01-30
Fix validator scripts utils

jjyr:
Note: the current version `validator_utils.h` assumes the first 32 bytes of backend-validator is `rollup_script_hash` @TheWaWaR


c/validator_utils.h
```diff
@@ -444,9 +602,8 @@ int _load_cancel_challenge_witness(gw_context_t *ctx) {
     mol_seg_t *script_seg = &pair->script_seg;
     if (i < scripts_size) {
       mol_seg_res_t seg_res = MolReader_ScriptVec_get(&scripts_seg, i);
-      uint8_t error_num = *(uint8_t *)(&seg_res);
-      if (error_num != MOL_OK) {
-        return -1;
+      if (seg_res.errno != MOL_OK) {
+        return GW_ERROR_INVALID_DATA;
```
TheWaWaR, line 606: Some macro in evmc.h or somewhere else will let the g++ compile treat `errno` as special pre-processor, so can't compile.

jjyr: Is it possible to avoid it in the included side? like diable the macro after include the evm.h?

https://stackoverflow.com/questions/1543736/how-do-i-temporarily-disable-a-macro-expansion-in-c-c

TheWaWaR: Let me have a try.

TheWaWaR: Works

TheWaWaR (approved):

PR #115, opened 2021-03-19
Impl StateDB

EthanYuan:
StateDB is based on the underlying kv storage. It can store all versions of keys and their values. This ablity is needed to record all the states of every account based on SMT (Sparse Merkle Tree).

The StateDB is impl by using a fixed-length version suffix appending the original key. It utilizes and encapsulates the range search capabilities of the underlying RocksDb iterator, which can easily locate the key-value pairs closest to the target version.


crates/store/src/state_db.rs
```diff
@@ -1,57 +1,87 @@
...
+    pub fn is_genesis_version(&self) -> bool {
+        self.block_hash.is_none() && self.tx_index.is_none()
     }
 }
 
-// TODO implement this
 pub struct StateDBTransaction {
     inner: StoreTransaction,
-    _version: StateDBVersion,
+    block_num: u64,
```
jjyr, line 52:
```suggestion
    block_number: u64,
```

keep a consistent naming style

crates/store/src/state_db.rs
```diff
@@ -86,6 +116,102 @@ impl StateDBTransaction {
         }
         Ok(())
     }
+
+    fn get_block_num_and_tx_index(
+        inner: &StoreTransaction,
```
jjyr, line 121:
```suggestion
        db: &StoreTransaction,
```

The name `inner` doesn't make sense here

crates/store/src/state_db.rs
```diff
@@ -86,6 +116,102 @@ impl StateDBTransaction {
         }
         Ok(())
     }
+
+    fn get_block_num_and_tx_index(
```
jjyr, line 120: Put this function outside the structure namespace.

crates/store/src/state_db.rs
```diff
@@ -86,6 +116,102 @@ impl StateDBTransaction {
...
+        inner: &StoreTransaction,
+        ver: &StateDBVersion,
+    ) -> Result<(u64, u32), Error> {
+        if ver.is_genesis_version() {
+            return Ok((BLOCK_NUMBER_ZERO, TX_INDEX_ZERO));
+        }
+        let block_hash = &ver
+            .block_hash
+            .ok_or_else(|| "Block hash doesn't exist".to_owned())?;
+        let block_num = inner
```
jjyr, line 130:
``` rust
let tx_index = version.tx_index.ok_or(|| "tx index is none".to_owned())?;
let block = db.get_block(block_hash).ok_or(|| "block doesn't exists".to_owned())?;
if block.transactions().len() <= tx_index as usize {
    return Err("invalid tx_index".to_owned());
}

Ok((block_number, tx_index))
```

jjyr: Pls keep the code simple and avoid using abbreviations.

crates/store/src/state_db.rs
```diff
@@ -86,6 +116,102 @@ impl StateDBTransaction {
...
+                } else {
+                    None
+                }
+            })?
+            .ok_or_else(|| "Tx index doesn't exist".to_owned())?;
+        Ok((block_num, tx_idx))
+    }
+
+    // This private constructor can be injected with mock data by unit test
+    fn from_tx_index(inner: StoreTransaction, block_num: u64, tx_index: u32) -> Self {
```
jjyr, line 157:
```suggestion
    #[cfg(test)]
    fn from_tx_index(inner: StoreTransaction, block_num: u64, tx_index: u32) -> Self {
```

jjyr: Only compile this function in testing.

crates/store/src/state_db.rs
```diff
@@ -86,6 +116,102 @@ impl StateDBTransaction {
...
+        StateDBTransaction {
+            inner,
+            block_num,
+            tx_index,
+        }
+    }
+
+    fn get_account_smt_root_and_count(&self) -> Result<(H256, u32), Error> {
```
jjyr, line 165:
```suggestion
    fn get_current_account_merkle_state(&self) -> Result<AccountMerkleState, Error> {
```

crates/store/src/state_db.rs
```diff
@@ -86,6 +116,102 @@ impl StateDBTransaction {
...
+            .get_block(&block_hash)?
+            .ok_or_else(|| "Block doesnt exist".to_owned())?;
+        let account_merkle_state = block.raw().post_account();
+        Ok((
+            account_merkle_state.merkle_root().unpack(),
+            account_merkle_state.count().unpack(),
+        ))
+    }
+
+    fn get_key_with_ver_sfx(&self, key: &[u8]) -> Vec<u8> {
```
jjyr, line 184:
```suggestion
    fn get_key_with_suffix(&self, key: &[u8]) -> Vec<u8> {
```

crates/store/src/state_db.rs
```diff
@@ -86,6 +116,102 @@ impl StateDBTransaction {
...
+        [
+            key,
+            &self.block_num.to_be_bytes(),
+            &self.tx_index.to_be_bytes(),
+        ]
+        .concat()
+    }
+
+    fn get_ori_key<'a>(&self, raw_key: &'a [u8]) -> &'a [u8] {
```
jjyr, line 193:
```suggestion
    fn get_original_key<'a>(&self, raw_key: &'a [u8]) -> &'a [u8] {
```

crates/store/src/state_db.rs
```diff
@@ -86,6 +116,102 @@ impl StateDBTransaction {
...
+    }
+
+    fn filter_value_of_seek(&self, ori_key: &[u8], raw_iter: &DBRawIterator) -> Option<Box<[u8]>> {
+        if !raw_iter.valid() {
+            return None;
+        }
+        match raw_iter.key() {
+            Some(raw_key_found) => {
+                if ori_key != self.get_ori_key(raw_key_found) {
+                    return None;
```
jjyr, line 204: Use `eprint!` to print an error.

crates/store/src/state_db.rs
```diff
@@ -196,15 +327,20 @@ impl<'a> CodeStore for StateTree<'a> {
...
     fn get_data(&self, data_hash: &H256) -> Option<Bytes> {
         match self.db.get(COLUMN_DATA, data_hash.as_slice()) {
             Some(slice) => Some(Bytes::from(slice.to_vec())),
             None => None,
         }
     }
 }
+
+#[cfg(test)]
+mod tests;
```
jjyr, line 346: We should put tests under `crates/store/src/tests` and define the tests module in the `crates/store/src/lib.rs`.

crates/store/src/state_db.rs
```diff
@@ -1,57 +1,87 @@
...
+        raw_iter.seek_for_prev(raw_key);
+        self.filter_value_of_seek(key, &raw_iter)
     }
 
+    // TODO: this trait method will be deleted in the future.
     fn get_iter(&self, col: Col, mode: IteratorMode) -> DBIter {
         self.inner.get_iter(col, mode)
     }
 
     fn insert_raw(&self, col: Col, key: &[u8], value: &[u8]) -> Result<(), Error> {
```
jjyr, line 69:
```
assert_ne!(value, FLAG_DELETE_VALUE, "forbid inserting the delete flag");
```

crates/store/src/transaction.rs
```diff
@@ -482,3 +485,6 @@ struct CustodianChange {
     sudt_script_hash: H256,
     amount: u128,
 }
+
+#[cfg(test)]
+mod tests;
```
jjyr, line 490: Remove this and define it in the right place.

jjyr (changes requested):

crates/store/src/state_db.rs
EthanYuan: This solution is ok and the file structure seems more clean. As the tests is defined as a module, which and the target module are at the same level, some build-in unit test functions like testing private are not be valid any more. Please note.

crates/store/src/state_db.rs
EthanYuan: Good idea.

crates/store/src/state_db.rs
```diff
@@ -45,17 +44,43 @@ impl StateDBVersion {
     pub fn is_genesis_version(&self) -> bool {
         self.block_hash.is_none() && self.tx_index.is_none()
     }
+
+    fn lock_block_number_and_tx_index(&self, db: &StoreTransaction) -> Result<(u64, u32), Error> {
```
jjyr, line 48:
```suggestion
    fn load_block_number_and_tx_index(&self, db: &StoreTransaction) -> Result<(u64, u32), Error> {
```

crates/store/src/state_db.rs
```diff
@@ -45,17 +44,43 @@ impl StateDBVersion {
...
+        if self.is_genesis_version() {
+            return Ok((0u64, 0u32));
+        }
+        let block_hash = &self
+            .block_hash
+            .ok_or_else(|| "Block hash doesn't exist".to_owned())?;
+        let block_number = db
```
jjyr, line 55: Get the number from block, don't read it from db.

crates/store/src/state_db.rs
```diff
@@ -45,17 +44,43 @@ impl StateDBVersion {
...
+            .get_block_number(block_hash)?
+            .ok_or_else(|| "Block number doesn't exist".to_owned())?;
+        let block = db
+            .get_block(block_hash)?
+            .ok_or_else(|| "Block doesn't exist".to_owned())?;
+        let tx_index = match self.tx_index {
+            Some(tx_index) => {
+                if block.transactions().len() <= tx_index as usize {
```
jjyr, line 63:
```suggestion
                if block.transactions().len() >= tx_index as usize {
```

jjyr (changes requested):

crates/store/src/state_db.rs
jjyr: Why we still pass the tests?

EthanYuan: State db's unit test's data are all be injected by the constructor from_tx_index, that means 
decoupling with the data of block, tx and other domain related.

This case (except equal case) is may not covered by all tests.

EthanYuan: I will try to make it up somewhere.

crates/store/src/tests/state_db.rs
```diff
@@ -0,0 +1,329 @@
...
+    assert_eq!(false, state_db_version.is_genesis_version());
+    assert!(
+        store.state_at(state_db_version).is_err(),
+        "block doesn't exist"
+    );
+
+    let state_db_version = StateDBVersion::from_tx_index(block_hash.into(), 1u32);
+    assert_eq!(false, state_db_version.is_genesis_version());
+    assert!(
+        store.state_at(state_db_version).is_err(),
```
jjyr, line 70:
```suggestion
        assert_eq!(store.state_at(state_db_version).unwrap_err(), ...);
```

crates/store/src/state_db.rs
```diff
@@ -1,57 +1,124 @@
...
+        let block_hash = &self
+            .block_hash
+            .ok_or_else(|| "Block hash doesn't exist".to_owned())?;
+        let block = db
+            .get_block(block_hash)?
+            .ok_or_else(|| "Block doesn't exist".to_owned())?;
+        let block_number = block.raw().number().unpack();
+        let tx_index = match self.tx_index {
+            Some(tx_index) => {
+                if tx_index as usize <= block.transactions().len() {
```
jjyr, line 64:
```suggestion
                if tx_index as usize < block.transactions().len() {
```

jjyr (changes requested):

jjyr (approved):

PR #121, opened 2021-04-01
Impl clear state db

EthanYuan:
When a block is detached due to some reason, the account state associated with the block needs to be cleared. This function is achieved by recording all the states of each block inserted by the state db.

After the confirmation number is reached, the block can no longer be detached. At this time, we delete the state record of the confirmed block. Therefore, the state record will always keep the size at a certain fixed level.


crates/db/src/db.rs
```diff
@@ -207,7 +207,10 @@ impl RocksDB {
 }
 
 #[inline]
-pub(crate) fn cf_handle(db: &OptimisticTransactionDB, col: Col) -> Result<&ColumnFamily> {
+pub(crate) fn cf_handle<'a>(
```
jjyr, line 210: The lifetime mark is useless.

crates/db/src/db.rs
```diff
@@ -207,7 +207,10 @@ impl RocksDB {
...
+    db: &'a OptimisticTransactionDB,
+    col: &str,
```
jjyr, line 212: We should update the definition of Col, not just replace the type.

```
type Col<'a> = &'a str;
```

jjyr (changes requested):
Please rebase this PR

crates/store/src/transaction.rs
```diff
@@ -477,8 +483,107 @@ impl StoreTransaction {
...
+
+    fn prune_block_state_record(&self, current_block_number: u64) -> Result<(), Error> {
+        if current_block_number <= NUMBER_OF_CONFIRMATION {
+            return Ok(());
+        }
+        let to_be_pruned_block_number = current_block_number - NUMBER_OF_CONFIRMATION - 1;
+        let block_hash = self.get_block_hash_by_number(to_be_pruned_block_number)?;
+        let block_hash = match block_hash {
+            Some(block_hash) => block_hash,
+            None => match to_be_pruned_block_number {
```
jjyr, line 509:
``` rust
Some(block_hash) => block_hash,
None if to_be_pruned_block_number == 0 => return Ok(()),
_ => return Err(...),
```

crates/store/src/transaction.rs
```diff
@@ -477,8 +483,107 @@ impl StoreTransaction {
...
+
+    fn clear_block_state_record(&self, block_hash: &H256) -> Result<(), Error> {
+        self.clear_block_state_and_record(block_hash, false)
+    }
+
+    fn clear_block_state(&self, block_hash: &H256) -> Result<(), Error> {
+        self.clear_block_state_and_record(block_hash, true)
+    }
+
+    fn clear_block_state_and_record(
```
jjyr, line 525: Change this function to `iter_block_state_record() -> Iterator<...>`.
And implement the delete logic in the `clear_block_state` & `clear_block_state_record`

crates/store/src/transaction.rs
```diff
@@ -477,8 +483,107 @@ impl StoreTransaction {
...
+            if need_clear_state {
+                let column = self.get_column_from_state_record(&key)?;
+                self.delete(column.as_str(), &value)?;
+            }
+            self.delete(COLUMN_BLOCK_STATE_RECORD, &key)?;
+        }
+        Ok(())
+    }
+
+    fn get_state_record_key(
```
jjyr, line 547: Do not implement the function on struct if it doesn't need to use the `self`.

crates/store/src/transaction.rs
```diff
@@ -477,8 +483,107 @@ impl StoreTransaction {
...
+        &self,
+        block_hash: &H256,
+        tx_index: u32,
+        col: Col,
+    ) -> Result<Vec<u8>, Error> {
+        let column =
+            u8::from_str_radix(col, 10).map_err(|_| "Parse column to int failed".to_owned())?;
```
jjyr, line 554: Why????

crates/store/src/transaction.rs
```diff
@@ -477,8 +483,107 @@ impl StoreTransaction {
...
+
+    fn get_column_from_state_record(&self, key: &[u8]) -> Result<String, Error> {
+        let column = key
+            .last()
+            .ok_or_else(|| "Decode column failed".to_owned())?;
+        Ok(column.to_string())
+    }
+
+    #[cfg(test)]
+    pub fn clear_block_account_state(
```
jjyr, line 571: Why????

jjyr (changes requested):
Pls make sure you are satisfied with the code before sending a PR.

jjyr (approved):

PR #126, opened 2021-04-16
BREAKING CHANGE: Support tron signature verification

zeroqn:
Add Secp256k1Tron to account lock manage
## Breaking Change
Add `tron_account_lock` to `programs` in scripts-deploy.json

### DIFF
--- a/crates/block-producer/src/runner.rs
+++ b/crates/block-producer/src/runner.rs
@@ -139,10 +139,6 @@ pub fn run(config: Config) -> Result<()> {
             eth_lock_script_type_hash.unpack(),
             Box::new(Secp256k1Eth::default()),
         );
-        account_lock_manage.register_lock_algorithm(
-            eth_lock_script_type_hash.unpack(),
-            Box::new(Secp256k1Tron::default()),
-        );
         Arc::new(Generator::new(
             backend_manage,
             account_lock_manage,
--- a/crates/block-producer/src/runner.rs
+++ b/crates/block-producer/src/runner.rs
@@ -9,7 +9,10 @@ use gw_chain::chain::Chain;
 use gw_common::H256;
 use gw_config::Config;
 use gw_generator::{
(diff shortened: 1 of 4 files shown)


crates/block-producer/src/runner.rs
```diff
@@ -136,6 +139,10 @@ pub fn run(config: Config) -> Result<()> {
             eth_lock_script_type_hash.unpack(),
             Box::new(Secp256k1Eth::default()),
         );
+        account_lock_manage.register_lock_algorithm(
+            eth_lock_script_type_hash.unpack(),
```
jjyr, line 143: We do not enable this lock in the current release, Pls remove this

crates/tools/src/deploy_genesis.rs
```diff
@@ -201,9 +201,10 @@ pub fn deploy_genesis(
         GwPack::pack(&deployment_result.polyjuice_validator.script_type_hash),
     ];
 
-    let mut allowed_eoa_type_hashes: Vec<gw_packed::Byte32> = vec![GwPack::pack(
-        &deployment_result.eth_account_lock.script_type_hash,
-    )];
+    let mut allowed_eoa_type_hashes: Vec<gw_packed::Byte32> = vec![
+        GwPack::pack(&deployment_result.eth_account_lock.script_type_hash),
+        GwPack::pack(&deployment_result.tron_account_lock.script_type_hash),
```
jjyr, line 206: Remove this line

jjyr (changes requested):

crates/block-producer/src/runner.rs
zeroqn: Done

crates/tools/src/deploy_genesis.rs
zeroqn: Done

jjyr (approved):

PR #130, opened 2021-04-22
Poa lock generator

jjyr:


crates/block-producer/src/rpc_client.rs
```diff
@@ -120,6 +143,120 @@ impl RPCClient {
...
+                    .await?,
+            )?;
+            cursor = Some(cells.last_cursor);
+            assert!(
+                cells.objects.len() <= 1,
+                "Never returns more than 1 identity cells"
+            );
+            cell = cells.objects.into_iter().find_map(|cell| {
+                // delete cells with data & type
+                if !cell.output_data.is_empty() || cell.output.type_.is_some() {
```
TheWaWaR, line 184: Why filter out cells with data and type? I think target identity cell must have data and type script.

TheWaWaR (changes requested):

crates/block-producer/src/rpc_client.rs
jjyr: You are right, this is a copy-paste error.

jjyr: I will close this PR please use @EthanYuan 's fork

PR #135, opened 2021-04-24
feat: Add web3 indexer

bitrocks:


crates/block-producer/src/main.rs
```diff
@@ -102,12 +104,37 @@ fn run() -> Result<()> {
...
+    // create web3 indexer
+    let web3_indexer = match config.web3_indexer {
+        Some(web3_indexer_config) => {
+            let pool = smol::block_on(async {
+                PgPoolOptions::new()
+                    .max_connections(5)
+                    .connect(&web3_indexer_config.database_url)
+                    .await
+            })?;
+            let l2_sudt_type_script_hash = web3_indexer_config.l2_sudt_type_script_hash;
```
jjyr, line 116:
```suggestion
            let l2_sudt_type_script_hash = rollup_config.l2_sudt_validator_script_type_hash();
```

crates/block-producer/src/main.rs
```diff
@@ -102,12 +104,37 @@ fn run() -> Result<()> {
...
+            let polyjuce_type_script_hash = web3_indexer_config.polyjuice_script_type_hash;
+            let rollup_type_hash = web3_indexer_config.rollup_type_hash;
```
jjyr, line 118:
```suggestion
            let rollup_type_hash = rollup_script_hash;
```

jjyr: We can read these two values from the current context.

crates/tools/src/main.rs
```diff
@@ -115,6 +115,13 @@ fn main() {
                         .required(true)
                         .help("Polyjuice binaries directory path"),
                 )
+                .arg(
+                    Arg::with_name("database-url")
+                        .short("d")
+                        .takes_value(true)
+                        .required(true)
```
jjyr, line 122: Make it optional

jjyr (changes requested):

crates/tools/src/main.rs
bitrocks: updated.

jjyr (approved):

PR #139, opened 2021-04-26
Fix infinite loop indexer rpc query

zeroqn:


crates/block-producer/src/rpc_client.rs
```diff
@@ -155,7 +155,13 @@ impl RPCClient {
                     )
                     .await?,
             )?;
+
+            if cells.last_cursor.is_empty() {
+                println!("no payment cell");
+                return Ok(collected_cells);
```
jjyr, line 161: We should return an error since we haven't collected enough capacity.

jjyr (changes requested):

crates/block-producer/src/rpc_client.rs
zeroqn: Ok

jjyr (approved):

PR #151, opened 2021-04-30
integrate withdrawal lock

zeroqn:
Generate withdrawal cells
Revert withdrawal cells

### DIFF
--- a/crates/block-producer/src/rpc_client.rs
+++ b/crates/block-producer/src/rpc_client.rs
@@ -800,6 +800,47 @@ impl RPCClient {
         Ok(collected)
     }
 
+    pub async fn query_rollup_config_cell(&self) -> Result<Option<CellInfo>> {
+        let search_key = SearchKey {
+            script: self.rollup_config_type_script.clone().into(),
+            script_type: ScriptType::Type,
+            filter: None,
+        };
+        let order = Order::Desc;
+        let limit = Uint32::from(1);
+
+        let mut cells: Pagination<Cell> = to_result(
+            self.indexer_client
+                .request(
+                    "get_cells",
+                    Some(ClientParams::Array(vec![
+                        json!(search_key),
+                        json!(order),
+                        json!(limit),
+                    ])),
+                )
+                .await?,
+        )?;
+        if let Some(cell) = cells.objects.pop() {
+            let out_point = {
+                let out_point: ckb_types::packed::OutPoint = cell.out_point.into();
+                OutPoint::new_unchecked(out_point.as_bytes())
+            };
+            let output = {
+                let output: ckb_types::packed::CellOutput = cell.output.into();
+                CellOutput::new_unchecked(output.as_bytes())
+            };
+            let data = cell.output_data.into_bytes();
+            let cell_info = CellInfo {
+                out_point,
+                output,
+                data,
+            };
+            return Ok(Some(cell_info));
+        }
+        Ok(None)
+    }
+
     pub async fn get_transaction(&self, tx_hash: H256) -> Result<Option<Transaction>> {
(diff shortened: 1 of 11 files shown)


crates/block-producer/src/rpc_client.rs
```diff
@@ -51,7 +51,7 @@ fn to_jsonh256(v: H256) -> JsonH256 {
     h.into()
 }
 
-fn to_result<T: DeserializeOwned>(output: Output) -> anyhow::Result<T> {
+pub fn to_result<T: DeserializeOwned>(output: Output) -> anyhow::Result<T> {
```
jjyr, line 54: This is a internal function, should not make it public

crates/block-producer/src/transaction_skeleton.rs
```diff
@@ -80,6 +84,13 @@ impl TransactionSkeleton {
     pub fn signature_entries(&self) -> Vec<SignatureEntry> {
         let mut entries: HashMap<[u8; 32], SignatureEntry> = Default::default();
         for (index, input) in self.inputs.iter().enumerate() {
+            // Skip withdrawal lock witness args
+            if let Some(witness_args) = self.witnesses().get(index) {
```
jjyr, line 88: Why do we have signed witnesses?

crates/block-producer/src/withdrawal.rs
```diff
@@ -0,0 +1,699 @@
...
+        CollectedCustodianCells {
+            cells_info: Default::default(),
+            capacity: 0,
+            sudt: Default::default(),
+            fullfilled_sudt_script: Default::default(),
+        }
+    }
+}
+
+async fn query_finalized_custodian_cells(
```
jjyr, line 444: Put this function into rpc_client.rs

jjyr (changes requested):

crates/block-producer/src/rpc_client.rs
zeroqn: How about pub(crate) or RPCClient::to_result ?
I place some cell query code only used by withdrawal in withdrawal.rs, it needs to access these code.

crates/block-producer/src/transaction_skeleton.rs
zeroqn: These are not signed withnesses, they are ```UnlockWithdrawalWitness``` and use to revert withdrawal cells back to custodian cells,

crates/block-producer/src/withdrawal.rs
zeroqn: Done

crates/block-producer/src/rpc_client.rs
zeroqn: Done

crates/block-producer/src/withdrawal.rs
zeroqn: Done

crates/block-producer/src/rpc_client.rs
```diff
@@ -89,11 +94,46 @@ fn parse_deposit_request(
...
+        }
+    }
+}
+
 #[derive(Clone)]
 pub struct RPCClient {
     pub indexer_client: HttpClient,
     pub ckb_client: HttpClient,
     pub rollup_type_script: ckb_types::packed::Script,
+    pub rollup_config_cell_dep: gw_types::packed::CellDep,
```
jjyr, line 136: This field is useless

zeroqn: Take ```rollup_config_cell_dep``` from ```BlockProducerConfig``` now. Done

jjyr (approved):

PR #152, opened 2021-04-30
Add load rollup config syscall

TheWaWaR:


jjyr:
This syscall exposes too many details to the user space, `RollupConfig` is an internal concept, the Godwoken backend script shouldn't access it directly. We should expose `check_sudt_account` or similar syscalls instead of this one.

crates/generator/src/syscalls/mod.rs
```diff
@@ -404,6 +405,12 @@ impl<'a, S: State, C: ChainStore, Mac: SupportMachine> Syscalls<Mac> for L2Sysca
                 machine.set_register(A0, Mac::REG::from_u8(SUCCESS));
                 Ok(true)
             }
+            SYS_LOAD_ROLLUP_CONFIG => {
```
jjyr, line 408: This syscall exposes too many details to the user space, RollupConfig is an internal concept, the Godwoken backend script shouldn't access it directly. We should expose check_sudt_account or similar syscalls instead of this one.

jjyr (changes requested):

crates/generator/src/syscalls/mod.rs
jjyr: I changed my mind. The backend is whitelisted, so they're under control.

jjyr (approved):

PR #161, opened 2021-05-07
Register secp256k1 data

TheWaWaR:
Register secp256k1 data for `ecrecover` pre-compiled EVM contract to use.

NOTE: this PR add a config field (`secp_data_dep`) in `BlockProducerConfig`, which is a **BREAKING CHANGE** in config, and user should re-generate the `config.toml` config file.

### DIFF
--- a/crates/config/src/config.rs
+++ b/crates/config/src/config.rs
@@ -43,6 +43,8 @@ pub struct GenesisConfig {
     pub rollup_type_hash: H256,
     pub meta_contract_validator_type_hash: H256,
     pub rollup_config: RollupConfig,
+    // For load secp data and use in challenge transaction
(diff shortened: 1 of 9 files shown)


crates/config/src/config.rs
```diff
@@ -60,6 +60,8 @@ pub struct BlockProducerConfig {
     pub stake_cell_lock_dep: CellDep,
     pub poa_lock_dep: CellDep,
     pub poa_state_dep: CellDep,
+    // For load secp data and use in challenge transaction
+    pub secp_data_dep: CellDep,
```
jjyr, line 64: Move this to GenesisConfig.

jjyr (changes requested):

jjyr (approved):

PR #182, opened 2021-05-14
feat: Support block parameter

bitrocks:
This PR add `block_number` parameter to JSON RPC `get_balance | get_storage_at | get_nonce | execute_raw_l2transaction`:

1. The response value would be wrapped as `option`, when `block_number` > `tip_block_number`, it just returns `None`;
2. The `execute_raw_l2transaction` impl is still not decoupled from mem_pool, as it needs  accessing `generator` to [execute_transaction](https://github.com/nervosnetwork/godwoken/pull/182/files#diff-99a766892812406083a959acd68f0fd2a6f62ff590f6187457f62038b499f76eR251);

### DIFF
--- a/crates/mem-pool/src/pool.rs
+++ b/crates/mem-pool/src/pool.rs
@@ -240,7 +240,10 @@ impl MemPool {
         block_hash: H256,
     ) -> Result<RunResult> {
         let db = self.store.begin_transaction();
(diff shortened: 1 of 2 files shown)


jjyr:
This PR will be reopen after modified.

PR #183, opened 2021-05-16
fix: withdrawal failure

zeroqn:
#### mempool
filter withdrawal request which capacity doesn't pass minimal capacity
#### block producer
don't package withdrawal request if no enough finalized custodian
invalid code hash because missing custodian type script cell dep


crates/block-producer/src/block_producer.rs
```diff
@@ -464,17 +556,18 @@ impl BlockProducer {
...
-            &self.rpc_client,
-        )
-        .await?;
-        tx_skeleton
-            .cell_deps_mut()
-            .extend(generated_withdrawal_cells.deps);
+            collected_custodians,
+        )?;
+        let mut resolved_deps =
+            resolve_type_deps(&self.rpc_client, &generated_withdrawal_cells.inputs).await?;
```
jjyr, line 566: Which type_deps do you expect to resolve? IMO there are only two deps: withdraw-lock & sUDT-type. We can pass these two from configure file.

crates/block-producer/src/block_producer.rs
```diff
@@ -483,17 +576,20 @@ impl BlockProducer {
...
         {
-            tx_skeleton
-                .cell_deps_mut()
-                .extend(reverted_withdrawals.deps);
+            let mut resolved_deps =
+                resolve_type_deps(&self.rpc_client, &reverted_withdrawals.inputs).await?;
+            resolved_deps.extend(tx_skeleton.cell_deps_mut().drain(..));
+            resolved_deps.extend(reverted_withdrawals.deps.drain(..));
+
+            *tx_skeleton.cell_deps_mut() = resolved_deps.into_iter().collect();
```
jjyr, line 592: Why not just call `tx_skeleton.cell_deps_mut().extend(...)`?

crates/block-producer/src/rpc_client.rs
```diff
@@ -680,7 +679,7 @@ impl RPCClient {
             )?;
 
             if cells.last_cursor.is_empty() {
-                return Err(anyhow!("no enough custodian cells"));
+                return Ok(collected);
```
jjyr, line 682: Why return Ok when we can't get enough custodian cells?

crates/block-producer/src/rpc_client.rs
```diff
@@ -765,6 +766,68 @@ impl RPCClient {
...
+                    .await?,
+            )?;
+
+            if cells.last_cursor.is_empty() {
+                return Ok(None);
+            }
+            cursor = Some(cells.last_cursor);
+
+            for cell in cells.objects.into_iter() {
+                let sudt_type_script = match cell.output.type_.clone() {
```
jjyr, line 813: You should also verify the lock.args which is not verified on-chain, an attacker can break a godwoken node by fill an invalid args. Pls check the `sanitize_deposit_cells` function.

crates/block-producer/src/withdrawal.rs
```diff
@@ -10,148 +9,209 @@ use gw_types::{
...
     time::{SystemTime, UNIX_EPOCH},
 };
 
-pub struct GeneratedWithdrawals {
-    pub deps: Vec<CellDep>,
-    pub inputs: Vec<InputCellInfo>,
-    pub outputs: Vec<(CellOutput, Bytes)>,
+#[derive(Debug)]
+pub struct AvailableCustodians {
+    pub capacity: u64,
```
jjyr, line 29: We should use u128 if it is an accumulated result.

crates/block-producer/src/withdrawal.rs
```diff
@@ -10,148 +9,209 @@ use gw_types::{
...
+
+        Generator {
+            rollup_context,
+            ckb_custodian,
+            sudt_custodians,
+            withdrawals: Default::default(),
+        }
+    }
 
+    pub fn insert(&mut self, req: &WithdrawalRequest, block: &L2Block) -> Result<()> {
```
jjyr, line 97: Rename to `push_and_verify` or something similar, the name `insert` does not reflect the purpose of this function.

jjyr (changes requested):

crates/block-producer/src/rpc_client.rs
zeroqn: The purpose of this function is changed, the result will be also be used to filter withdrawal requests.
Even if the total balance isn't enough, we can still process some of them. For example, A withdraw 200, B withdraw 400.
But we only have 300 finalized available at the time, we can still process A request and skip B.

crates/block-producer/src/rpc_client.rs
zeroqn: The result script is only used to calculate estimate withdrawal cell's capacity in ```minimal_capacity_verifier```. I think it's better move it to ```minimal_capacity_verifier()```.

crates/block-producer/src/block_producer.rs
zeroqn: This would be a breaking change to ```config.toml```. I think we can add ```ckb-indexer-url``` to ```gw-tools``` to generate cell dep for user.

crates/block-producer/src/withdrawal.rs
zeroqn: rename to ```include_and_verify```.
Done

crates/block-producer/src/withdrawal.rs
zeroqn: Done

crates/block-producer/src/rpc_client.rs
zeroqn: Now it's verified. Done

crates/block-producer/src/rpc_client.rs
zeroqn: Since we gather finalized custodians info from db, restore this function.

crates/block-producer/src/block_producer.rs
zeroqn: Done

crates/block-producer/src/block_producer.rs
zeroqn: Done. Now user should set ```l1_sudt_tyep_dep``` in ```config.toml```.

crates/block-producer/src/withdrawal.rs
```diff
@@ -11,149 +13,307 @@ use gw_types::{
...
+    sync::Arc,
     time::{SystemTime, UNIX_EPOCH},
 };
 
-pub struct GeneratedWithdrawals {
-    pub deps: Vec<CellDep>,
-    pub inputs: Vec<InputCellInfo>,
-    pub outputs: Vec<(CellOutput, Bytes)>,
+lazy_static::lazy_static! {
+    static ref CUSTODIAN_TYPE_SCRIPTS: Arc<Mutex<HashMap<[u8; 32], Script>>> = Arc::new(Mutex::new(Default::default()));
```
jjyr, line 31: Do not use global static mutable variable, put this into the block-producer structure.

jjyr (changes requested):

crates/block-producer/src/withdrawal.rs
zeroqn: Done

crates/store/src/transaction.rs
```diff
@@ -395,25 +399,48 @@ impl StoreTransaction {
...
-            });
-        let withdrawal_assets = block.withdrawals().into_iter().map(|withdrawal| {
-            let raw = withdrawal.raw();
-            CustodianChange {
-                sudt_script_hash: raw.sudt_script_hash().unpack(),
-                amount: raw.amount().unpack(),
-                capacity: raw.capacity().unpack(),
-            }
-        });
-        self.update_custodian_assets(deposit_assets, withdrawal_assets)?;
```
jjyr, line 416: Rename to update_finalized_custodian_assets

crates/block-producer/src/block_producer.rs
```diff
@@ -259,6 +259,67 @@ impl BlockProducer {
...
+                        }),
+                        Err(err) => {
+                            log::debug!("get custodian type script err {}", err);
+                            None
+                        }
+                    }
+                });
+
+                let to_custodian = has_script.filter_map(|(hash, script)| {
+                    match db.get_custodian_asset(hash.into()) {
```
jjyr, line 297: Pls rename to `get_finalized_custodian_assets`.

crates/block-producer/src/withdrawal.rs
```diff
@@ -11,149 +13,301 @@ use gw_types::{
...
+    ) -> Self {
+        let mut total_sudt_capacity = 0u128;
+        let mut sudt_custodians = HashMap::new();
+
+        for (sudt_type_hash, (balance, type_script)) in available_custodians.sudt.into_iter() {
+            let (change, _data) =
+                generate_finalized_custodian(rollup_context, balance, type_script.clone());
+            let change_capacity: u64 = change.capacity().unpack();
+            total_sudt_capacity = total_sudt_capacity.saturating_add(change_capacity as u128);
+            sudt_custodians.insert(sudt_type_hash, (change_capacity, balance, type_script));
```
jjyr, line 72: Please use structs instead of using tuples.

jjyr (changes requested):

crates/block-producer/src/block_producer.rs
```diff
@@ -464,23 +526,30 @@ impl BlockProducer {
...
-            .extend(generated_withdrawal_cells.outputs);
+        .await?
+        {
+            tx_skeleton
+                .cell_deps_mut()
+                .extend(generated_withdrawal_cells.deps);
+            tx_skeleton
+                .inputs_mut()
+                .extend(generated_withdrawal_cells.inputs);
+            println!(
```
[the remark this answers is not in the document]
zeroqn: Done

crates/block-producer/src/withdrawal.rs
```diff
@@ -11,149 +13,324 @@ use gw_types::{
...
+            ckb_custodian,
+            sudt_custodians,
+            withdrawals: Default::default(),
+        }
+    }
 
+    pub fn verified_output(
+        &self,
+        req: &WithdrawalRequest,
+        block: &L2Block,
```
jjyr, line 121: Can we remove this?

jjyr (approved):

jjyr:
Pls cherry-pick this PR to master

PR #208, opened 2021-06-07
Adjust for interoperability

jjyr:
Changes:

1. Remove compatible_chain_id from RollupConfig
2. Introduces a new concept: short_address, which is a variable lengthed value. Currently, we only support 20 bytes short_address.
3. Support new syscall to recover EoA script from the signature.


zeroqn (approved):
LGTM

PR #217, opened 2021-06-17
Add withdraw cli

classicalliu:


crates/tools/src/main.rs
```diff
@@ -229,6 +230,84 @@ fn main() {
                         .takes_value(true)
                         .default_value("http://127.0.0.1:8119")
                         .help("Godwoken jsonrpc rpc sever URL"),
+                )
+                .arg(
```
jjyr, line 234: IMO the cli should automatically try `gw_` version RPC after we failed to request a RPC.

```
Failed to request /xxx
Retry /gw_xxx ...
```

jjyr (changes requested):

crates/tools/src/main.rs
classicalliu: Updated.

jjyr (approved):

PR #219, opened 2021-06-18
Update generate-config command

EthanYuan:
Update OPTIONS( delete one and add new three):	
    ...
    ~~-p  \<polyjuice-binaries-dir-path\>   Polyjuice binaries directory path~~	
    -k  \<privkey-path\>			            The private key file path
    -c  \<scripts-deploy-config-path\> 	    Scripts depoly config json file path
    -u  \<rpc-server-url\>				    Rpc server url [default: localhost:8119]

The original command line is as follows:

```bash
cargo run --bin gw-tools -- generate-config -g deploy/genesis-deploy-result.json -s deploy/scripts-deploy-result.json  
-o config.toml -p deploy/
```

The updated command line is as follows:

```bash
cargo run --bin gw-tools -- generate-config -g deploy/genesis-deploy-result.json -s deploy/scripts-deploy-result.json  
-o deploy/node1/config.toml -k deploy/node1/pk -c deploy/scripts-deploy.json 
```

All fields of the output config.toml will be filled in correctly. 

Note: 
- The scripts-deploy.json is generated by the command prepare-scripts.
- Each call can only output config file for one node.

### DIFF
--- a/crates/tools/src/generate_config.rs
+++ b/crates/tools/src/generate_config.rs
@@ -8,14 +8,31 @@ use gw_config::{
     RPCServerConfig, StoreConfig, WalletConfig, Web3IndexerConfig,
 };
 use gw_jsonrpc_types::godwoken::L2BlockCommittedInfo;
-use std::{fs, path::Path};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 
-const BACKEND_BINARIES_DIR: &str = "godwoken-scripts/c/build";
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Debug)]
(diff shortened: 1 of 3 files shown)


crates/tools/src/generate_config.rs
```diff
@@ -51,13 +76,42 @@ pub fn generate_config(
...
+        hex::decode_to_slice(
+            node_wallet_info
+                .block_assembler_code_hash
+                .strip_prefix("0x")
+                .expect("get code hash"),
+            &mut hash as &mut [u8],
+        )?;
+        hash
+    };
+    let args: [u8; 20] = {
```
jjyr, line 91: The args does not has to be 20 bytes. We should use Bytes to represent args.

jjyr (changes requested):

jjyr (approved):

PR #224, opened 2021-06-22
feat: implement test mode

zeroqn:
### brief
This pr add ```test mode```. In test mode, godwoken will stop produce new block and expose following jsonrpc methods

- tests_produce_block
- tests_should_produce_block
- tests_get_global_state

To produce an new block, we make a jsonrpc to ```tests_produce_block```.

### config change

```toml
[test_mode]
enable = false
```

Add above entry to ```config.toml```

#### method tests_produce_block
Update on-chain global state. Right now only produce new block is supported. 

##### method tests_should_produce_block
Whether this node has right to produce next block

##### method tests_get_global_state
Return on-chain global state

#### Example

```sh
echo '{
    "id": 2,
    "jsonrpc": "2.0",
    "method": "tests_produce_block",
    "params": ["none"]
}' \
| tr -d '\n' \
| curl -H 'content-type: application/json' -d @- \
http://localhost:8119//

echo '{
    "id": 2,
    "jsonrpc": "2.0",
    "method": "tests_should_produce_block",
    "params": []
}' \
| tr -d '\n' \
| curl -H 'content-type: application/json' -d @- \
http://localhost:8119

echo '{
    "id": 2,
    "jsonrpc": "2.0",
    "method": "tests_get_global_state",
    "params": []
}' \
| tr -d '\n' \
| curl -H 'content-type: application/json' -d @- \
http://localhost:8119/
```


crates/config/src/config.rs
```diff
@@ -87,3 +88,31 @@ pub struct Web3IndexerConfig {
...
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum TestMode {
+    Enable,
+    Disable,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct TestModeConfig {
```
jjyr, line 100: We can use `TestMode` in the config. Please remove this.

crates/rpc-server/src/registry.rs
```diff
@@ -40,14 +51,27 @@ pub struct Registry {
     generator: Arc<Generator>,
     mem_pool: Arc<MemPool>,
     store: Store,
+    test_mode: TestMode,
+    tests_rpc_impl: Arc<BoxedTestsRPCImpl>,
```
jjyr, line 55: IMO we do not need a trait here, we can use TestModeControl type.

jjyr (changes requested):

crates/rpc-server/src/registry.rs
zeroqn: But this result in cyclic dependency. ```TestModeControl``` needs ```poa``` and ```rpc-client``` in ```block-producer```.

jjyr (approved):

PR #262, opened 2021-07-18
I fix: insert data_hash key into SMT

jjyr:


TheWaWaR (dismissed):

TheWaWaR:
Please fix the commit message: `I fix: insert data_hash key into SMT`

PR #266, opened 2021-07-18
fix(challenger): cancel challenge may spend owner cell twice

zeroqn:
Cancel challenge should filter inputs from verifier tx.


jjyr (dismissed):

jjyr:
Pls rebase

jjyr (approved):

PR #274, opened 2021-07-22
Update rpc get_block, return L2BlockWithStatus.

EthanYuan:


crates/rpc-server/src/registry.rs
```diff
@@ -128,14 +128,26 @@ async fn ping() -> Result<String> {
...
-) -> Result<Option<L2BlockView>> {
+) -> Result<Option<L2BlockWithStatus>> {
     let block_hash = to_h256(block_hash);
     let db = store.begin_transaction();
     let block_opt = db.get_block(&block_hash)?.map(|block| {
         let block_view: L2BlockView = block.into();
         block_view
     });
-    Ok(block_opt)
+    let l2_block_status = if H256::zero() != db.reverted_block_smt()?.get(&block_hash)? {
```
jjyr, line 138:
```suggestion
    let l2_block_status = if !db.reverted_block_smt()?.get(&block_hash)?.is_zero() {
```

jjyr (changes requested):

jjyr:
https://github.com/nervosnetwork/godwoken/pull/276 is merged

crates/rpc-server/src/registry.rs
```diff
@@ -131,14 +135,27 @@ async fn ping() -> Result<String> {
...
-        block_view
-    });
-    Ok(block_opt)
+    let block_opt = db.get_block(&block_hash)?;
+    let mut status = L2BlockStatus::Unfinalized;
+    if let Some(ref block) = block_opt {
+        if !db.reverted_block_smt()?.get(&block_hash)?.is_zero() {
+            status = L2BlockStatus::Reverted;
+        } else {
+            let tip_block_number = db.get_last_valid_tip_block()?.raw().number().unpack();
```
jjyr, line 148: We also need to check that block is exist in the block_smt & return None if it is not found.

jjyr (changes requested):

crates/rpc-server/src/registry.rs
```diff
@@ -131,14 +135,30 @@ async fn ping() -> Result<String> {
...
-    let block_opt = db.get_block(&block_hash)?.map(|block| {
-        let block_view: L2BlockView = block.into();
-        block_view
-    });
-    Ok(block_opt)
+    let block_opt = db.get_block(&block_hash)?;
+    let mut status = L2BlockStatus::Unfinalized;
+    if let Some(ref block) = block_opt {
+        if db.block_smt()?.get(&block.smt_key().into())?.is_zero() {
+            return Ok(None);
```
jjyr, line 146: If a block is reverted, we can't find it from block_smt, which means this RPC would never return a reverted block.

jjyr (changes requested):

crates/rpc-server/src/registry.rs
zeroqn: Block's smt key is block number. For reverted block, we should check reverted block smt first. If valid tip block number is smaller than queried reverted block, block smt will return H256::zero. If valid tip block number is equal or bigger than reverted block, block smt will return different block hash.

jjyr (dismissed):

jjyr (approved):

zeroqn (approved):

Flouse (approved):

PR #284, opened 2021-07-29
fix: cancel polyjuice tx challenge failure

zeroqn:
Save SYS_LOAD_ACCOUNT_SCRIPT script into RunResult, we can then
package these scripts into cancel challenge witness.

### DIFF
--- a/crates/generator/src/syscalls/mod.rs
+++ b/crates/generator/src/syscalls/mod.rs
@@ -508,12 +508,19 @@ impl<'a, S: State, C: ChainStore> L2Syscalls<'a, S, C> {
     fn set_account_count(&mut self, count: u32) {
         self.result.account_count = Some(count);
     }
-    fn get_script(&self, script_hash: &H256) -> Option<Script> {
-        self.result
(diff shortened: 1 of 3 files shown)


crates/generator/src/syscalls/mod.rs
```diff
@@ -508,12 +508,19 @@ impl<'a, S: State, C: ChainStore> L2Syscalls<'a, S, C> {
...
+        let opt_script = self
+            .result
             .new_scripts
             .get(script_hash)
             .map(|data| Script::from_slice(&data).expect("Script"))
-            .or_else(|| self.code_store.get_script(&script_hash))
+            .or_else(|| self.code_store.get_script(&script_hash));
+
+        if let Some(ref script) = opt_script {
+            self.result.get_scripts.insert(script.as_slice().to_vec());
```
jjyr, line 520: We should save Script instead of Vec<u8>, script is a molecule type which shares references, so we can save memory.

jjyr (approved):

jjyr (approved):

PR #294, opened 2021-08-04
feat(block producer): santize deposit capacity base on custodian output

zeroqn:


crates/types/Cargo.toml
```diff
@@ -16,3 +16,4 @@ cfg-if = "0.1"
 gw-hash = { path = "../hash", default-features = false }
 ckb-fixed-hash = { version = "0.38.0", optional = true }
 sparse-merkle-tree = { version = "0.5.2-rc1", default-features = false }
+ckb-types = { version = "0.38.0", default-features = false }
```
jjyr, line 19: Set optional = true, and add this feature to std

crates/types/src/extension.rs
```diff
@@ -143,3 +143,12 @@ impl packed::RollupConfig {
         self.as_reader().hash()
     }
 }
+
+impl packed::CellOutput {
```
jjyr, line 147: move it to offchain.rs

jjyr (changes requested):

jjyr (approved):

PR #295, opened 2021-08-04
feat(block producer): sanitize withdrawal base on output cell capacity

zeroqn:


crates/types/src/offchain.rs
```diff
@@ -28,3 +28,9 @@ impl CellOutput {
             .map(|c| c.as_u64())
     }
 }
+
+impl std::hash::Hash for Script {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.as_reader().hash().hash(state)
```
jjyr, line 34: self.as_reader().hash(state)

jjyr (approved):

PR #299, opened 2021-08-09
fix(challenger): cancel polyjuice tx call contract challenge failure v2

zeroqn:
Pass sys_load_data through cell_deps

### DIFF
--- a/crates/block-producer/src/debugger.rs
+++ b/crates/block-producer/src/debugger.rs
@@ -16,7 +16,7 @@ use gw_types::{
     prelude::*,
 };
 
-use crate::rpc_client::RPCClient;
+use crate::{rpc_client::RPCClient, types::TxStatus};
 
 pub async fn dump_transaction<P: AsRef<Path>>(
     dir: P,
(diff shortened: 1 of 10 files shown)


Flouse (approved):

crates/block-producer/src/debugger.rs
```diff
@@ -65,10 +65,11 @@ pub async fn build_mock_transaction(
             return Ok(vec![]);
         }
         // parse dep group
-        let cell = rpc_client
-            .get_cell(dep.out_point())
-            .await?
-            .ok_or_else(|| anyhow!("can't find dep group cell"))?;
+        let cell = match rpc_client.get_cell(dep.out_point()).await? {
+            Some(cell) => Some(cell),
+            None => rpc_client.get_cell_from_mempool(dep.out_point()).await?,
```
jjyr, line 70: Do we need to implement this in the rpc layer? If get_cell return null, we try to query mem pool, what do you think?

jjyr (approved):

PR #300, opened 2021-08-11
feat(metrics): add metrics crate and block metrics

magicalne:
This pr involves metrics collection with opentelemetry.

By enabling metrics collection:

1. Building godwoken with:

```shell
cargo build --features tokio --manifest-path crates/block-producer/Cargo.toml
```
2. Adding mectics block in config.toml:

```toml
[metrics]
endpoint = "http://host:port"
worker_threads = 2
```

Screenshot from 2021-08-11 20-08-53
Screenshot from 2021-08-11 20-08-47


crates/block-producer/Cargo.toml
```diff
@@ -44,6 +49,7 @@ serde_json = "1.0"
 smol = "1.2.5"
 lazy_static = "1.4"
 sqlx = { version = "0.5", features = [ "runtime-async-std-native-tls", "postgres", "sqlite", "chrono", "decimal" ] }
+openssl = { version = "0.10.29", features = ["vendored"] }
```
jjyr, line 52: why we need this crate?

crates/block-producer/src/runner.rs
```diff
@@ -158,6 +158,11 @@ async fn poll_loop(
 }
 
 pub fn run(config: Config, skip_config_check: bool) -> Result<()> {
+    // Enable metrics config.
+    #[cfg(feature = "tokio")]
```
jjyr, line 162: why checking tokio feature here? all the things should be hidden behind the gw_metrics crate

jjyr (changes requested):

crates/block-producer/Cargo.toml
magicalne: That's a mistake.:(

crates/block-producer/src/runner.rs
magicalne: I thought that way before. But that means I need to check gw_metrics feature when i add a new metric every time, which is weird as well..
Instead, I set tokio as an optional dependency and reexport tokio feature to upper level.
Or... I can put metrics utils into a new crate. So crates can always depend on metric-utils and optionall depend on metric-config. Sounds better?
Need to figure out it's feasible or not.

crates/block-producer/Cargo.toml
magicalne: I just check my local build and it's not a mistake. On my fedora34, cannot build without it. That's due to a cross-platform compilation on [openssl](https://github.com/launchbadge/sqlx/issues/473#issuecomment-655517309).

crates/block-producer/src/runner.rs
magicalne: As we discussed offline, we have two features: `default` and `metrics` on metrics crate.

PR #316, opened 2021-08-27
feat(rpc): debug dump cancel challenge tx

zeroqn:
- feat(rpc): add ```debug_dump_cancel_challenge_tx``` method
- chore(config): add ```enable_debug_rpc``` to ```debug``` in ```config.toml```
- feat(tools): add ```dump-cancel-challenge-tx``` command to ```gw-tools```


zeroqn:
Set default field value ```false``` for ```enable_debug_rpc```. It should not panic now.

PR #327, opened 2021-09-08
fix: store generator incorrect state db sub state

zeroqn:


crates/store/src/state_db.rs
```diff
@@ -168,6 +168,24 @@ impl CheckPoint {
         db: &StoreTransaction,
         db_mode: StateDBMode,
     ) -> Result<(u64, u32), Error> {
+        let block_offset = |withdrawal_count: u32, tx_count: u32| -> u32 {
+            if 0 == withdrawal_count {
```
jjyr, line 172: We can replace the `if` statement with
`(withdrawal_count + 1 + tx_count).saturating_sub(1)`

crates/store/src/state_db.rs
```diff
@@ -168,6 +168,24 @@ impl CheckPoint {
...
+                tx_count // 0 is prev txs
+            } else {
+                // For example: 2 withdrawals + 2 txs, we should have 5 checkpoint, a.k.a 0..=4
+                withdrawal_count + 1 + tx_count.saturating_sub(1)
+            }
+        };
+        let tx_offset = |withdrawal_count: u32, tx_index: u32| -> u32 {
+            if 0 == withdrawal_count {
```
jjyr, line 180: The if statement is equals to:
```
withdrawal_count + 1 + tx_index
```

jjyr (approved):

jjyr:
We should write a test case for this.

PR #339, opened 2021-09-14
feat: validate sudt proxy contract creator by adding whitelist in [rpc] config

magicalne:
## new config [rpc]

```toml
[rpc]
allowed_sudt_proxy_creator_account_id = [1] ## only account_id = 1 can create sudt proxy contract
sudt_proxy_code_hashes = ["0x43a008ec973b648bd71ad67e6b66f2be8a6fa88e89c7dad046c948b00aa866aa"] ## supported sudt proxy contract code hashes
```

### DIFF
--- a/crates/generator/src/account_whitelist.rs
+++ b/crates/generator/src/account_whitelist.rs
@@ -1,0 +1,53 @@
+use gw_common::H256;
+use gw_types::offchain::RunResult;
+use log::debug;
+
+pub struct SUDTProxyAccountWhitelist {
+    allowed_sudt_proxy_creator_account_id: Vec<u32>,
+    sudt_proxy_code_hashes: Vec<H256>,
+}
+
+impl SUDTProxyAccountWhitelist {
+    pub fn new(
+        allowed_sudt_proxy_creator_account_id: Vec<u32>,
+        sudt_proxy_code_hashes: Vec<H256>,
+    ) -> Self {
+        Self {
+            allowed_sudt_proxy_creator_account_id,
+            sudt_proxy_code_hashes,
+        }
+    }
+
+    /// Only accounts in white list could create sUDT proxy contract.
+    pub fn validate(&self, run_result: &RunResult, from_id: u32) -> bool {
+        for k in run_result.write_data.keys() {
+            debug!(
+                "whiltelist: from_id: {:?}, code_hash: {:?}",
+                &from_id,
+                hex::encode(k.as_slice())
+            );
+
+            // Contract create syscall stores code in write_data.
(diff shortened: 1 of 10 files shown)


crates/generator/src/account_whitelist.rs
```diff
@@ -0,0 +1,53 @@
...
+    ) -> Self {
+        Self {
+            allowed_sudt_proxy_creator_account_id,
+            sudt_proxy_code_hashes,
+        }
+    }
+
+    /// Only accounts in white list could create sUDT proxy contract.
+    pub fn validate(&self, run_result: &RunResult, from_id: u32) -> bool {
+        for k in run_result.write_data.keys() {
```
jjyr, line 23: We have three shortcuts:

1. if from_id in the whitelist, we could immediately return true.
2. if run_result.scripts.is_empty, we could also immediately return true since there are no new account been created.
3. if the to account's script.code_hash isn't polyjuice, we can return true.

otherwise we should read code_hash from state tree, instead of iterate all keys.

jjyr (changes requested):

crates/generator/src/account_whitelist.rs
jjyr: Please rename this module to `erc20_creator_whitelist.rs`

crates/generator/src/erc20_creator_whitelist.rs
```diff
@@ -0,0 +1,64 @@
...
+                "whiltelist: from_id: {:?}, code_hash: {:?}",
+                &from_id,
+                hex::encode(k.as_slice())
+            );
+
+            // Contract create syscall stores code in write_data.
+            // check code hash is sudt proxy contract
+            if self.sudt_proxy_code_hashes.contains(k) {
+                // check account id from whitelist
+                if !self
```
jjyr, line 45: Remove this since we already check whitelist at the beginning

crates/generator/src/erc20_creator_whitelist.rs
```diff
@@ -0,0 +1,64 @@
...
+    ) -> Self {
+        Self {
+            allowed_sudt_proxy_creator_account_id,
+            sudt_proxy_code_hashes,
+        }
+    }
+
+    /// Only accounts in white list could create sUDT proxy contract.
+    pub fn validate(&self, run_result: &RunResult, from_id: u32) -> bool {
+        if self
```
jjyr, line 23: Please add another shortcut, if whitelist is empty return true.

jjyr (changes requested):

jjyr (approved):

PR #353, opened 2021-09-23
feat: add node info api

magicalne:
Node info API will return godwoken version and backend configs.

```sh
$ curl http://localhost:8119 -X POST -H "Content-Type: application/json" -d '{"jsonrpc": "2.0", "method":"gw_get_node_info", "params": [], "id": 1}' 
```
The response looks like:
```json
{
  "jsonrpc": "2.0",
  "result": {
    "version": "0.6.4 145deca",
    "backends": [
      {
        "validator_code_hash": "0x8b00c1699f34524ddd5fe27c92a696fb63160ec63fafbbecd003cd659d88642e",
        "generator_code_hash": "0xef5effea165330db9bf6cee52fb8c42c44479e3a5e0c9d3972eb84e948734382",
        "validator_script_type_hash": "0xa16fc0d4cc190f560090f1a195113808b0128d2c2c4e7a7331e384b5b3f23d9d"
      },
      {
        "validator_code_hash": "0x5f29228f6b53741b6ba42a90fc18b32228fcbd74ee2cdc7f1cb6bad3b0a27f93",
        "generator_code_hash": "0xcbb3ea2f404cede64b72b6e3a1b4826aa04cefaf18f790e9d009ce0c604eeec7",
        "validator_script_type_hash": "0xdf17edcdcfc2635639186895d71d0e453cb5cb510252fd7fbec1d25858666605"
      },
      {
        "validator_code_hash": "0xbe4da16f86f199c7a4b73e06be9732e384b528c7072abca7a00cfaa4551114d1",
        "generator_code_hash": "0x2822eb72c552e1f25219ba27866f831200aeaa6dbb6d89688ce69cb25f9eb443",
        "validator_script_type_hash": "0xccff2ab8be3e794edd03e2f310e34bff95f404bb3c971e2c2a3c8d5056047f8b"
      }
    ]
  },
  "id": 1
}
```

### DIFF
--- a/crates/common/src/backend.rs
+++ b/crates/common/src/backend.rs
@@ -1,8 +1,0 @@
-use sparse_merkle_tree::H256;
-
-#[derive(Clone)]
-pub struct BackendInfo {
-    pub validator_code_hash: H256,
-    pub generator_code_hash: H256,
-    pub validator_script_type_hash: H256,
-}
--- a/crates/common/src/backend.rs
+++ b/crates/common/src/backend.rs
@@ -1,0 +1,8 @@
+use sparse_merkle_tree::H256;
+
+#[derive(Clone)]
+pub struct BackendInfo {
+    pub validator_code_hash: H256,
+    pub generator_code_hash: H256,
+    pub validator_script_type_hash: H256,
+}
--- a/crates/generator/src/backend_manage.rs
+++ b/crates/generator/src/backend_manage.rs
@@ -11,20 +11,20 @@ pub struct Backend {
     pub validator_script_type_hash: H256,
 }
 
-impl From<Backend> for BackendInfo {
-    fn from(backend: Backend) -> Self {
+impl Backend {
+    fn get_backend_info(&self) -> BackendInfo {
         let mut validator_code_hash = [0u8; 32];
         let mut hasher = new_blake2b();
-        hasher.update(&backend.validator);
+        hasher.update(&self.validator);
         hasher.finalize(&mut validator_code_hash);
         let mut generator_code_hash = [0u8; 32];
         let mut hasher = new_blake2b();
-        hasher.update(&backend.generator);
+        hasher.update(&self.generator);
         hasher.finalize(&mut generator_code_hash);
         BackendInfo {
             validator_code_hash: validator_code_hash.into(),
             generator_code_hash: generator_code_hash.into(),
-            validator_script_type_hash: backend.validator_script_type_hash,
+            validator_script_type_hash: self.validator_script_type_hash,
         }
     }
 }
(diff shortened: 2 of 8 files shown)


crates/common/src/backend.rs
```diff
@@ -0,0 +1,8 @@
+use sparse_merkle_tree::H256;
```
jjyr, line 1: The gw_jsonrpc_types::godwoken::BackendInfo is enough for use, we can remove this module

crates/generator/src/backend_manage.rs
```diff
@@ -11,6 +11,24 @@ pub struct Backend {
     pub validator_script_type_hash: H256,
 }
 
+impl From<Backend> for BackendInfo {
+    fn from(backend: Backend) -> Self {
```
jjyr, line 15: The `From` & `Into` normally implies light operation, but we calculate blake2b for a large binary input. I suggest we do not implement From for this conversion.

crates/generator/src/backend_manage.rs
```diff
@@ -58,4 +76,12 @@ impl BackendManage {
     pub fn get_backend(&self, code_hash: &H256) -> Option<&Backend> {
         self.backends.get(code_hash)
     }
+
+    pub fn get_backend_info(&self) -> Vec<BackendInfo> {
```
jjyr, line 80: We only need BackendInfo in rpc module.

crates/generator/src/generator.rs
```diff
@@ -775,6 +775,10 @@ impl Generator {
 
         Ok((output, data))
     }
+
+    pub fn get_backend_info(&self) -> Vec<gw_common::backend::BackendInfo> {
```
jjyr, line 779: The get_backend function is enough, we can call get_backend in the RPC module and convert it to BackendInfo.

crates/rpc-server/src/registry.rs
```diff
@@ -726,6 +728,18 @@ async fn compute_l2_sudt_script_hash(
     Ok(to_jsonh256(l2_sudt_script.hash().into()))
 }
 
+async fn get_node_info(generator: Data<Generator>) -> Result<NodeInfo> {
+    let backend = generator
+        .get_backend_info()
+        .into_iter()
+        .map(|backend_info| backend_info.into())
```
jjyr, line 735: We calculate blake2b for three binaries every time, which isn't necessary.

jjyr (changes requested):

crates/common/src/backend.rs
magicalne: This is due to two impl of H256.
gw_jsonrpc_types::godwoken uses ckb_fixed_hash::H256 only and the others use smt::H256. And the common crate is used by most crates, so I think this is a good place to make value to value transformations ?

crates/rpc-server/src/registry.rs
magicalne: Let me put this in the struct of `Generator` as a field of backend_info: Vec\<BeckendInfo\>.

jjyr (approved):

PR #357, opened 2021-09-26
feat(mempool): customize number of mem block withdrawals and txs

zeroqn:


crates/block-producer/src/block_producer.rs
```diff
@@ -244,83 +249,114 @@ impl BlockProducer {
...
+                        err
+                    );
+                    self.mem_pool.lock().await.reset_mem_block()?;
+                    return Err(err);
+                }
+            };
+
+            if tx.as_slice().len() > MAX_BLOCK_BYTES as usize {
+                // Drop some percentage of withdrawals and txs
+                let max_withdrawals = block_withdrawals
```
jjyr, line 331: We should try to drop txs first, then only to drop withdrawals when there are no tx get packaged into the block.

And we can adjust the strategy to drop half of the txs(or withdrawals) each time.

crates/block-producer/src/block_producer.rs
```diff
@@ -244,83 +249,114 @@ impl BlockProducer {
...
         }
 
-        // get txs & withdrawal requests from mem pool
-        let (opt_finalized_custodians, block_param) = {
-            let mem_pool = self.mem_pool.lock().await;
-            mem_pool.output_mem_block()?
-        };
-        let deposit_cells = block_param.deposits.clone();
+        let mut mem_block_output_param = OutputParam::default();
+        loop {
```
jjyr, line 253: we should define a max retries count, then raise error after meet the condition.

crates/config/src/config.rs
```diff
@@ -86,9 +86,26 @@ pub struct ChallengerConfig {
...
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct BlockCooldown {
+    pub withdrawals: usize,
+    pub txs: usize,
+}
+
+impl Default for BlockCooldown {
+    fn default() -> Self {
+        BlockCooldown {
+            withdrawals: 10, // drop 10% of withdrawals
```
jjyr, line 98: I think we can just try to drop half txs to simplify our logic.

crates/mem-pool/src/pool.rs
```diff
@@ -49,6 +49,41 @@ use crate::{
...
 };
 
+#[derive(Debug)]
+pub struct MemBlockLimit {
+    pub max_withdrawals: usize,
+    pub max_txs: usize,
+}
+
+impl MemBlockLimit {
+    pub fn set(&mut self, max_withdrawals: usize, max_txs: usize) {
```
jjyr, line 59: Can we define this structure as inmmutable?

jjyr (changes requested):

crates/block-producer/src/block_producer.rs
zeroqn: Fixed

crates/block-producer/src/block_producer.rs
zeroqn: Add max retry count.
Now we max try count check and withdrawals/txs zero check.

crates/config/src/config.rs
zeroqn: Fixed

crates/block-producer/src/block_producer.rs
```diff
@@ -244,83 +249,124 @@ impl BlockProducer {
...
+                        "[produce_next_block] Failed to composite submitting transaction: {}",
+                        err
+                    );
+                    self.mem_pool.lock().await.reset_mem_block()?;
+                    return Err(err);
+                }
+            };
+
+            if tx.as_slice().len() > MAX_BLOCK_BYTES as usize {
+                max_retry -= 1;
```
jjyr, line 331: Use `while max_retry > 0 {}` is more intuitive than `loop {}`

crates/config/src/config.rs
```diff
@@ -86,9 +86,28 @@ pub struct ChallengerConfig {
...
+pub struct BlockCooldown {
+    pub withdrawals: usize,
+    pub txs: usize,
+    pub max_retry: usize,
+}
+
+impl Default for BlockCooldown {
+    fn default() -> Self {
+        BlockCooldown {
+            withdrawals: 10, // drop 10% of withdrawals
```
jjyr, line 99:
```suggestion
            withdrawals: 50, // drop 50% of withdrawals
```

crates/mem-pool/src/pool.rs
```diff
@@ -341,8 +376,21 @@ impl MemPool {
...
+
+        let (mem_block, post_merkle_state) = if output_param.retry_count > 0 {
+            self.repackage_mem_block(output_param)?
+        } else {
+            let mem_block = self.mem_block.clone();
+            let mem_db_state = self.fetch_state_db(&db)?;
+
+            (mem_block, mem_db_state.state_tree()?.merkle_state()?)
+        };
+
```
jjyr, line 393: Since we do not need to call `self.mem_block`, we can refactoring the code below into another function?

crates/mem-pool/src/pool.rs
```diff
@@ -429,18 +468,148 @@ impl MemPool {
...
+        assert!(repackage_block.state_checkpoints().is_empty());
+        assert!(repackage_block.withdrawals().is_empty());
+        assert!(repackage_block.finalized_custodians().is_none());
+        assert!(repackage_block.deposits().is_empty());
+        assert!(repackage_block.txs().is_empty());
+
+        let mut state =
+            state_db.state_tree_with_merkle_state(mem_block.prev_merkle_state().to_owned())?;
+
+        // NOTE: Must have at least one tx to have correct psot block state
```
jjyr, line 525:
```suggestion
        // NOTE: Must have at least one tx to have correct post block state
```

crates/mem-pool/src/pool.rs
```diff
@@ -429,18 +468,148 @@ impl MemPool {
...
+
+            let deposit_cells = mem_block.deposits().to_vec();
+            let prev_state_checkpoint = mem_block
+                .txs_prev_state_checkpoint()
+                .ok_or_else(|| anyhow!("repackage mem block but no prev state checkpoint"))?;
+            repackage_block.push_deposits(deposit_cells, prev_state_checkpoint);
+
+            repackage_block.append_touched_keys(mem_block.touched_keys().clone().into_iter());
+        } else {
+            assert_eq!(tx_hashes.len(), 0, "reduce txs first");
```
jjyr, line 547:
```suggestion
            assert_eq!(tx_hashes.len(), 0, "must drop txs first");
```

crates/store/src/state_db.rs
```diff
@@ -271,7 +271,11 @@ impl CheckPoint {
                     unreachable!()
                 }
             },
-            SubState::MemBlock => Ok((MEMORY_BLOCK_NUMBER, 0)),
+            SubState::MemBlock => match db_mode {
+                StateDBMode::Genesis => Ok((MEMORY_BLOCK_NUMBER, 0)),
```
jjyr, line 275: Remove this case or raise a error because it is not actually exist

jjyr (changes requested):

crates/mem-pool/src/pool.rs
```diff
@@ -429,18 +460,154 @@ impl MemPool {
...
+        log::info!("[mem-pool] package mem block, retry count {}", retry_count);
+
+        let mem_block = &self.mem_block;
+        let state_db = self.fetch_state_db_with_mode(&db, MemBlockDBMode::Package)?;
+
+        let (withdrawal_hashes, deposits, tx_hashes) = {
+            let total =
+                mem_block.withdrawals().len() + mem_block.deposits().len() + mem_block.txs().len();
+            // Drop base on retry count
+            let mut remain = total / (output_param.retry_count + 1);
```
jjyr, line 496: remain must at least be 1. we should set it to 1 if it equals to 0.

jjyr (changes requested):

jjyr (approved):

Issue #347, opened 2021-09-20
Support block_hash / block_number parameter in execute_l2transaction RPC method [EIP-1898]


Kuzirashi:
Motivation:
- Support EIP https://eips.ethereum.org/EIPS/eip-1898
- Make it possible for godwoken-web3 to implement EIP-1898
- Increased compatibility with Geth
- Support for https://thegraph.com/ without forking

Related issue: https://github.com/RetricSu/godwoken-kicker/issues/107

I see in code that execute transaction already supports passing block info. The only thing needed to be changed is parsing this data from POST call.

cc @jjyr

Kuzirashi:
It turns out godwoken-web3 is using execute_raw_transaction instead of execute_transaction and execute_raw_transaction already supports passing block number.

Kuzirashi:
Yes, closing for now. I think godwoken has good enough support.

PR #362, opened 2021-09-30
Support passing optional block_number in execute_l2transaction

Kuzirashi:
This is my attempt at fixing #347. For a start it supports passing block number. No block hash support yet.

I have started godwoken-kicker with this configuration and it seems like I can deploy contract and call read/set methods just fine.


Kuzirashi:
It turns out godwoken-web3 is using execute_raw_transaction instead of execute_transaction and execute_raw_transaction already supports passing block number.

PR #377, opened 2021-10-20
feat(rpc): error tx receipt

zeroqn:
- for ```execute_l2transaction``` and ```execute_raw_l2transaction```, response error receipt in jsonrpc error data field
- for ```submit_l2transaction```, save error receipt to web3 indexer database and clear them after expiration blocks

### DIFF
--- a/crates/block-producer/src/runner.rs
+++ b/crates/block-producer/src/runner.rs
@@ -20,7 +20,9 @@ use gw_generator::{
     genesis::init_genesis,
     Generator,
 };
-use gw_mem_pool::{default_provider::DefaultMemPoolProvider, pool::MemPool};
+use gw_mem_pool::{
+    default_provider::DefaultMemPoolProvider, pool::MemPool, traits::MemPoolErrorTxHandler,
+};
 use gw_poa::PoA;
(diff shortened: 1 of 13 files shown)


crates/block-producer/src/runner.rs
```diff
@@ -423,11 +428,31 @@ pub fn run(config: Config, skip_config_check: bool) -> Result<()> {
...
+            let pg_pool = {
+                let config = config.web3_indexer.as_ref();
+                let init_pool = config.map(|web3_indexer_config| {
+                    smol::block_on(async {
+                        let mut opts: PgConnectOptions =
+                            web3_indexer_config.database_url.parse()?;
+                        opts.log_statements(log::LevelFilter::Debug)
+                            .log_slow_statements(log::LevelFilter::Warn, Duration::from_secs(5));
+                        PgPoolOptions::new()
+                            .max_connections(5)
```
Flouse, line 440: I am not sure if max_connections(5) is enough.

Flouse: I think it is sufficient if there are few time-consuming connections.
So, let's collect the `slow_statements` logs and make a decision later.

PR #428, opened 2021-11-23
Godwken benchmark

magicalne:
This pr contains such changes:

- Turns gw-tools to lib to share common code.
- Add gw-benchmark as a new crate bin.

One thing to mention:
Since the original godwoken_rpc is a blocking client and **blocking** reqwest client shouldn't be used in the context of "async". A blocking client can easily get into a deadlock in a tokio app. So I added an async version of the godwoken RPC client in the benchmark crate.


crates/tools/src/godwoken_rpc.rs
```diff
@@ -72,7 +72,7 @@ impl GodwokenRpcClient {
         short_address: JsonBytes,
     ) -> Result<Option<H256>> {
         let params = serde_json::to_value((short_address,))?;
-
+        dbg!("get_script_hash_by_short_address");
```
jjyr, line 75: Remove dbg before send PR

jjyr (changes requested):

jjyr (dismissed):

jjyr:
Is this PR ready for merge?

crates/benchmark/src/config.rs
```diff
@@ -0,0 +1,31 @@
...
+    fn default() -> Self {
+        Self {
+            interval: 1000,
+            batch: 10,
+            timeout: 120,
+            account_path: PathBuf::from("./accounts"),
+            gw_rpc_url: String::from("http://localhost:8119"),
+            polyman_url: String::from("http://localhost:6102"),
+            scripts_deploy_path: PathBuf::from("./scripts_deploy_results.json"),
+            rollup_type_hash: String::from("0x"),
```
Flouse, line 27: Next version, maybe we could read `rollup_type_hash` from `gw-config.toml` and remove this config from here.

Flouse (dismissed):

crates/benchmark/src/config.rs
magicalne: Let's just pass the parameters that we actually need. If we could use more configs from `config.toml`, then I'd like to involve `config.toml` instead.

jjyr:
Please cherry-pick this PR to develop branch.

jjyr:
Please cherry-pick this PR to develop branch and open a new PR

PR #431, opened 2021-11-26
chore: increase search range of deposits

jjyr:


crates/rpc-client/src/rpc_client.rs
```diff
@@ -393,7 +393,7 @@ impl RPCClient {
     /// return all lived deposit requests
     /// NOTICE the returned cells may contains invalid cells.
     pub async fn query_deposit_cells(&self, count: usize) -> Result<Vec<DepositInfo>> {
-        const BLOCKS_TO_SEARCH: u64 = 100;
+        const BLOCKS_TO_SEARCH: u64 = 2000;
```
Flouse, line 396: why is 2000 blocks?

jjyr: https://github.com/nervosnetwork/godwoken/blob/master/crates/mem-pool/src/constants.rs#L2

The max value of deposits in a block is 50 due to the cycles limitation(>600 M cycles).

This arguments affects how many history l1 block we need to search, if the BLOCK_TO_SEARCH == 100, means we can handle 100 * 50 deposits. But this is a idealy situation, if the deposits is more than this value in a short period, we may loss to handle some deposits (user can unlock these deposits by themselves), so we need to increase this value to improve the capacity of deposits.

Flouse (approved):

zeroqn (approved):

PR #447, opened 2021-12-01
feat: defragment finalize custodians

zeroqn:


crates/rpc-client/src/rpc_client.rs
```diff
@@ -728,6 +730,277 @@ impl RPCClient {
...
+                    // Invalid custodian type script
+                    let l1_sudt_script_type_hash =
+                        rollup_context.rollup_config.l1_sudt_script_type_hash();
+                    if sudt_type_script.code_hash() != l1_sudt_script_type_hash
+                        || sudt_type_script.hash_type() != ScriptHashType::Type.into()
+                    {
+                        continue;
+                    }
+
+                    if sudt_type_script.hash() != CKB_SUDT_SCRIPT_ARGS
```
jjyr, line 831: Please explain this line, why sudt_type_script should equals to CKB_SUDT_SCRIPT_ARGS?

crates/rpc-client/src/rpc_client.rs
```diff
@@ -728,6 +730,277 @@ impl RPCClient {
...
+                        log::error!("invalid sudt amount, out_point: {:?}", cell.out_point);
+                        continue;
+                    }
+                }
+
+                let (sudt_amount, type_hash) = match opt_sudt_type_script.as_ref() {
+                    Some(h) if h.hash() != CKB_SUDT_SCRIPT_ARGS => {
+                        (parse_sudt_amount(&cell).expect("sudt amount"), h.hash())
+                    }
+                    _ => (0, CKB_SUDT_SCRIPT_ARGS),
```
jjyr, line 843: IMO we should simply skip these custodian cells which hash equals to CKB_SUDT_SCRIPT_ARGS.

crates/rpc-client/src/rpc_client.rs
```diff
@@ -728,6 +730,277 @@ impl RPCClient {
...
+                };
+
+                // We'll try to cache up to max_custodian_cells different sudt custodians
+                if type_hash != CKB_SUDT_SCRIPT_ARGS
+                    && !withdrawals_amount.sudt.contains_key(&type_hash)
```
jjyr, line 848: So we only merge simple UDT custodian cells when someone withdraw it?

crates/rpc-client/src/rpc_client.rs
```diff
@@ -728,6 +730,277 @@ impl RPCClient {
...
+                        .or_insert_with(CandidateCustodians::<Reverse<_>>::default),
+                };
+
+                candidate_cells += 1;
+                candidate_capacity =
+                    candidate_capacity.saturating_add(custodian_cell.capacity as u128);
+                custodians.push(type_hash, Reverse(custodian_cell));
+
+                // Drop minimal amount/capacity custodian
+                if custodians.cells.len() > max_custodian_cells {
```
jjyr, line 877:
```
If custodians.cells.len() < max_custodian_cells{
                candidate_cells += 1;
                candidate_capacity =
                candidate_capacity.saturating_add(custodian_cell.capacity as u128);
                custodians.push(type_hash, Reverse(custodian_cell));
}
```

crates/rpc-client/src/rpc_client.rs
```diff
@@ -728,6 +730,277 @@ impl RPCClient {
...
+                    let min = custodians.pop().expect("minimal custodian");
+                    candidate_cells -= 1;
+                    candidate_capacity = candidate_capacity.saturating_sub(min.capacity as u128);
+                }
+
+                // Already fulfilled
+                if custodians.fulfilled || custodians.type_hash == CKB_SUDT_SCRIPT_ARGS {
```
jjyr, line 884: Should we check this before invoke `custodians.push`

crates/rpc-client/src/rpc_client.rs
zeroqn: Fixed. Now use enum to represent custodian type.

crates/rpc-client/src/rpc_client.rs
zeroqn: Fixed.

crates/rpc-client/src/rpc_client.rs
zeroqn: No. We can still cache them, if possible, combine them later.

crates/rpc-client/src/rpc_client.rs
zeroqn: If we get a bigger amount/capacity custodian, we should use it and drop smaller one.

crates/rpc-client/src/rpc_client.rs
```diff
@@ -728,6 +730,295 @@ impl RPCClient {
...
+                collected.capacity = collected.capacity.saturating_add(cell.capacity as u128);
+                collected.cells_info.push(cell.info.clone());
+
+                let (collected_amount, _) = {
+                    let sudt = collected.sudt.entry(custodians.type_hash.raw());
+                    sudt.or_insert((0, cell.info.output.type_().to_opt().unwrap_or_default()))
+                };
+                *collected_amount = collected_amount
+                    .checked_add(cell.amount)
+                    .expect("already check overflow");
```
jjyr, line 960: We should skip to collect this cell, this panic can be used by attackers.

crates/rpc-client/src/rpc_client.rs
```diff
@@ -1280,3 +1571,174 @@ impl RPCClient {
...
+impl CustodianTypeHash {
+    fn is_ckb(&self) -> bool {
+        matches!(self, CustodianTypeHash::Ckb)
+    }
+
+    fn is_sudt(&self) -> bool {
+        matches!(self, CustodianTypeHash::Sudt(_))
+    }
+
+    fn raw(&self) -> [u8; 32] {
```
jjyr, line 1590:
```suggestion
    fn sudt_type_hash(&self) -> Option<[u8; 32]> {
```

We should return Option, let the caller to call result.expect("no sudt type hash")

crates/rpc-client/src/rpc_client.rs
```diff
@@ -1280,3 +1571,174 @@ impl RPCClient {
...
+            self.amount.checked_add(amount).ok_or(AmountOverflow)?
+        };
+        self.capacity = {
+            let capacity = reverse_cell.0.capacity as u128;
+            self.capacity.saturating_add(capacity)
+        };
+
+        self.type_hash = type_hash;
+        self.cells.push(reverse_cell);
+        self.cell_len = self.cells.len();
```
jjyr, line 1659: Why don't call cells.len when we need?

PR #465, opened 2021-12-07
Use Jemalloc

Flouse:


crates/block-producer/src/main.rs
```diff
@@ -1,3 +1,10 @@
+#[cfg(not(target_env = "msvc"))]
+use tikv_jemallocator::Jemalloc;
+
+#[cfg(not(target_env = "msvc"))]
```
jjyr:
```suggestion
#[cfg(all(not(target_env = "msvc"), not(target_os = "macos")))]
```

https://github.com/nervosnetwork/ckb/blob/b3da297eda0b4bea7892b12124c783ce42bd701f/src/main.rs#L5

crates/block-producer/Cargo.toml
```diff
@@ -33,6 +33,7 @@ ckb-crypto = "0.100.0"
 ckb-fixed-hash = "0.100.0"
 ckb-types = "0.100.0"
 ckb-chain-spec = "0.100.0"
+tikv-jemallocator = { version = "0.4.0", features = ["unprefixed_malloc_on_supported_platforms", "profiling"] }
```
jjyr: `[target.'cfg(all(not(target_env = "msvc"), not(target_os="macos")))'.dependencies]`

https://github.com/nervosnetwork/ckb/blob/3c66fdbd8795841dce10c759c9d27e220340bb22/Cargo.toml#L99

jjyr: And do not enable profiling by default, it affect the performance.

We can use a feature to switch it.

https://github.com/nervosnetwork/ckb/blob/3c66fdbd8795841dce10c759c9d27e220340bb22/Cargo.toml#L107

jjyr (approved):

PR #468, opened 2021-12-08
Fix chain revert bad block

zeroqn:


crates/chain/src/chain.rs
```diff
@@ -361,16 +362,38 @@ impl Chain {
                         return Ok(SyncEvent::BadBlock { context });
                     }
 
-                    if let Some(_challenge_target) = self.process_block(
+                    println!("deposit_requests {}", deposit_requests.len());
```
jjyr: remove? or change to log

jjyr (dismissed):

crates/chain/src/chain.rs
zeroqn: typo

PR #479, opened 2021-12-10
fix: fix web3 missing indexed blocks on start

jjyr:


crates/web3-indexer/src/indexer.rs
```diff
@@ -61,7 +61,26 @@ impl Web3Indexer {
         }
     }
 
-    pub async fn store_genesis(&self, store: Store) -> Result<()> {
+    pub async fn fix_missing_blocks(&self, store: &Store) -> Result<()> {
+        let tip = store.get_tip_block()?;
```
zeroqn: valid tip?

jjyr: good catch!

jjyr: fixed

zeroqn (approved):

PR #498, opened 2021-12-15
Sync mem block from full node to readonly nodes.

magicalne:


crates/config/src/config.rs
```diff
@@ -214,11 +214,25 @@ impl Default for OffChainValidatorConfig {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct PublishMemBlockConfig {
+    pub hosts: Vec<String>,
+    pub topic: String,
+}
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct SubscribeMemBlockConfig {
```
jjyr:
```suggestion
pub struct SubscribeMemPoolConfig {
```

crates/config/src/config.rs
```diff
@@ -214,11 +214,25 @@ impl Default for OffChainValidatorConfig {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct PublishMemBlockConfig {
```
jjyr:
```suggestion
pub struct PublishMemPoolConfig {
```

crates/config/src/config.rs
```diff
@@ -214,11 +214,25 @@ impl Default for OffChainValidatorConfig {
...
+    pub topic: String,
+    pub group: String,
+}
+
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct MemPoolConfig {
     pub execute_l2tx_max_cycles: u64,
     #[serde(default = "default_restore_path")]
     pub restore_path: PathBuf,
+    pub pub_sync_mem_block: Option<PublishMemBlockConfig>,
```
jjyr:
```suggestion
    pub publish: Option<PublishMemPoolConfig>,
```

crates/config/src/config.rs
```diff
@@ -214,11 +214,25 @@ impl Default for OffChainValidatorConfig {
...
+    pub sub_sync_mem_block: Option<SubscribeMemBlockConfig>,
```
jjyr:
```suggestion
    pub subscribe: Option<SubscribeMemPoolConfig>,
```

crates/mem-pool/src/pool.rs
```diff
@@ -87,6 +90,9 @@ pub struct MemPool {
     pending_deposits: Vec<DepositInfo>,
     /// Mem block save and restore
     restore_manager: RestoreManager,
+    // Fan-out mem block from full node to readonly node
+    fan_out_mem_block_handler: Option<FanOutMemBlockHandler>,
```
jjyr: MemPoolPublishService is a more intuitive name for this.

crates/mem-pool/src/pool.rs
```diff
@@ -106,8 +112,9 @@ impl MemPool {
         generator: Arc<Generator>,
         provider: Box<dyn MemPoolProvider + Send>,
         error_tx_handler: Option<Box<dyn MemPoolErrorTxHandler + Send>>,
-        config: MemPoolConfig,
+        config: (MemPoolConfig, NodeMode),
```
jjyr: Why use tuple?, we can just add another parameter node_mode: NodeMode.

crates/mem-pool/src/sync/fan_in.rs
```diff
@@ -0,0 +1,69 @@
...
+use gw_config::SubscribeMemBlockConfig;
+use gw_types::packed::*;
+use gw_types::prelude::Unpack;
+use smol::lock::Mutex;
+
+use crate::pool::MemPool;
+
+use super::mq::{gw_kafka, Consume};
+
+pub(crate) struct FanInMemBlock {
```
jjyr: SubscribeMemPoolService, the module name is also non-intuitive.

crates/mem-pool/src/sync/fan_out.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub(crate) fn new(producer: gw_kafka::Producer) -> Self {
+        let (sender, receiver) = smol::channel::bounded(CHANNEL_BUFFER_SIZE);
+
+        let actor = FanOutMemBlockActor::new(receiver, producer);
+        smol::spawn(fan_out_handle(actor)).detach();
+        Self { sender }
+    }
+
+    pub(crate) fn new_tx(&self, tx: L2Transaction) {
+        let _ = smol::block_on(
```
jjyr: Output the error.

crates/mem-pool/src/sync/fan_out.rs
```diff
@@ -0,0 +1,76 @@
...
+        block_info: BlockInfo,
+    ) {
+        let next_mem_block = NextMemBlock::new_builder()
+            .block_info(block_info)
+            .withdrawals(withdrawals.pack())
+            .deposits(deposits.pack())
+            .txs(txs.pack())
+            .build();
+        let msg = RefreshMemBlockMessageUnion::NextMemBlock(next_mem_block);
+        let _ = smol::block_on(self.sender.send(msg));
```
jjyr: Output the error

docs/sync_mem_block.md
```diff
@@ -0,0 +1,36 @@
...
+topic = 'sync-mem-block'
+```
+
+## enalbe subscribing mem block on readonly node
+
+config.toml:
+
+```toml
+[mem_pool.sub_sync_mem_block]
+hosts = ['localhost:9092']
```
jjyr: What happend if we setup multiple hosts?

crates/mem-pool/src/pool.rs
```diff
@@ -1275,6 +1311,45 @@ impl MemPool {
...
+    // Only **ReadOnly** node needs this.
+    // Refresh mem block with those params.
+    pub(crate) fn refresh_mem_block(
+        &mut self,
+        block_info: BlockInfo,
+        withdrawals: Vec<WithdrawalRequest>,
+        deposits: Vec<DepositInfo>,
+        txs: Vec<L2Transaction>,
+    ) -> Result<()> {
+        let mem_block = MemBlock::new(block_info, self.mem_block.prev_merkle_state().to_owned());
```
jjyr: The second block should be the local `tip_block.post_merkle_state`.

jjyr: And we should ignore this update if the block_info isn't match the local block. hint: The local tip block should be parent block of `block_info`

crates/mem-pool/src/sync/fan_out.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn handle(&mut self, msg: RefreshMemBlockMessageUnion) {
+        if let Err(err) = self.producer.send(msg) {
+            log::error!("[Fan out mem block] message failed: {:?}", err);
+        }
+    }
+}
+
+async fn fan_out_handle(mut actor: FanOutMemBlockActor) {
+    log::info!("Fanout handle is started.");
+    while let Ok(msg) = actor.receiver.recv().await {
```
jjyr: 1. we should output the error from recv. Do not ignore any errors.
2. we should exit the actor / task if channel is ended, you can detect that from the error.

crates/mem-pool/src/sync/fan_out.rs
```diff
@@ -0,0 +1,76 @@
...
+pub(crate) struct FanOutMemBlockHandler {
+    sender: Sender<RefreshMemBlockMessageUnion>,
+}
+
+impl FanOutMemBlockHandler {
+    pub(crate) fn new(producer: gw_kafka::Producer) -> Self {
+        let (sender, receiver) = smol::channel::bounded(CHANNEL_BUFFER_SIZE);
+
+        let actor = FanOutMemBlockActor::new(receiver, producer);
+        smol::spawn(fan_out_handle(actor)).detach();
```
jjyr: Rename this function to start. The name `new` implies the function shouldn't has any side-effect

crates/mem-pool/src/sync/mq/gw_kafka.rs
```diff
@@ -0,0 +1,130 @@
...
+
+pub(crate) struct Producer {
+    producer: kafka::producer::Producer,
+    topic: String,
+}
+
+impl Producer {
+    pub(crate) fn new(hosts: Vec<String>, topic: String) -> Result<Self> {
+        let mut client = kafka::client::KafkaClient::new(hosts);
+        client.load_metadata(&[&topic]).map_err(|err| {
```
jjyr: Rename this function to connect. The name new implies the function shouldn't has any side-effect

crates/mem-pool/src/sync/mq/gw_kafka.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+impl Consumer {
+    pub(crate) fn new(
+        hosts: Vec<String>,
+        topic: String,
+        group: String,
+        fan_in: FanInMemBlock,
+    ) -> Result<Self> {
+        let consumer = kafka::consumer::Consumer::from_hosts(hosts)
```
jjyr: Rename this function to start. The name new implies the function shouldn't has any side-effect

jjyr (changes requested):

docs/sync_mem_block.md
magicalne: Hosts are the registered brokers for High-Availablity or Fail-Over in zookeeper. So we can achieve HA if we set multiple hosts.

crates/mem-pool/src/pool.rs
magicalne: Both are from the config. Plus if we add one more parameter, it would be **7** parameters in total, which triggers the "too many parameters warning". Maybe I can refactor the params by wrapping some of them into a new struct in another pr?

zeroqn: IMO, we dont need NodeMode here.
```NodeMode``` is only used in two places: ```push_transaction_with_db``` to skip ```verify_tx()``` and ```notify_new_tip```.

Why not directly call ```finalize_tx()``` instead? and for ```notify_new_tip```, just pass ```None``` to chain in runner.rs.

File: crates/mem-pool/src/pool.rs [hunk 16]
```diff
@@ -939,10 +964,21 @@ impl MemPool {
...
         let deposit_cells = self.pending_deposits.clone();
-        self.finalize_deposits(db, deposit_cells)?;
+        self.finalize_deposits(db, deposit_cells.clone())?;
+
+        // Fan-out next mem block to readonly node
+        if let Some(handler) = &self.fan_out_mem_block_handler {
+            handler.next_mem_block(
+                withdrawals,
+                deposit_cells,
+                txs.clone().collect(),
```
zeroqn: Transactions here are not finalized in 'new' mem block, and you broadcast them twice.
They get broadcasted again in ```self.push_transaction_with_db()``` later.
Maybe this cause nonce issue during syncing.

magicalne: Nice catch! txs in next_mem_block should be removed.

crates/mem-pool/src/sync/fan_out.rs
```diff
@@ -0,0 +1,76 @@
...
+        Self { sender }
+    }
+
+    pub(crate) fn new_tx(&self, tx: L2Transaction) {
+        let _ = smol::block_on(
+            self.sender
+                .send(RefreshMemBlockMessageUnion::L2Transaction(tx)),
```
zeroqn: try_send?

crates/mem-pool/src/sync/mq/gw_kafka.rs
```diff
@@ -0,0 +1,130 @@
...
+        let mut client = kafka::client::KafkaClient::new(hosts);
+        client.load_metadata(&[&topic]).map_err(|err| {
+            anyhow!(
+                "topic {:?} not found, load metadata error: {:?}",
+                &topic,
+                err
+            )
+        })?;
+        let producer = kafka::producer::Producer::from_client(client)
+            .with_ack_timeout(Duration::from_secs(1))
```
zeroqn: 1 second? I'm not sure if it works under pressure.

crates/mem-pool/src/sync/mq/gw_kafka.rs
```diff
@@ -0,0 +1,130 @@
...
+            .map_err(|err| anyhow!("Create consumer with error: {:?}", err))?;
+        Ok(Self { consumer, fan_in })
+    }
+}
+
+impl Consume for Consumer {
+    fn poll(&mut self) -> Result<()> {
+        let msg_sets = self
+            .consumer
+            .poll()
```
zeroqn: Is this ```poll()``` blocking?

magicalne: Not blocking. This task is spawned in smol.

crates/mem-pool/src/sync/mq/gw_kafka.rs
magicalne: I checked this again. This should be fine.

crates/mem-pool/src/sync/fan_out.rs
magicalne: Then how can we handle the "channel is full" error if we use try_send here. We cannot discard it. And we cannot block the thread.
I think we have to guarantee the producer will send messages fast enough.

crates/mem-pool/src/sync/mq/gw_kafka.rs
zeroqn: I mean interval poll instead of simply loop, avoid 100% cpu.

crates/mem-pool/src/sync/fan_out.rs
zeroqn: There's ```MAX_MEM_BLOCK_TXS```, we can use a slight bigger size for this channel.

zeroqn: In case kafka goes down, we can save tx in a VecDeque  in ```FanOutMemBlockActor```.

magicalne: We cannot do this. Kafka cluster has higher Availability than godwoken, apparently. It has quorum that we don't. We should not backup data for it.
Instead, kafka should be monitored very well.

zeroqn: Ok, simply log error if channel is full is also acceptable.

zeroqn (dismissed):

crates/mem-pool/src/pool.rs
```diff
@@ -1275,6 +1301,63 @@ impl MemPool {
...
+        );
+
+        Ok(Some(next_block_number))
+    }
+
+    // Only **ReadOnly** node needs this.
+    // Sync tx from fullnode to readonly.
+    pub(crate) fn append_tx(&mut self, tx: L2Transaction) -> Result<()> {
+        let db = self.store.begin_transaction();
+        let _ = self.finalize_tx(&db, tx)?;
```
jjyr: `let _ =` is unnecessary

crates/mem-pool/src/pool.rs
```diff
@@ -1275,6 +1301,63 @@ impl MemPool {
...
+
+        let post_merkle_state = tip_block.raw().post_account();
+        let mem_block = MemBlock::new(block_info, post_merkle_state);
+        self.mem_block = mem_block;
+
+        let db = self.store.begin_transaction();
+
+        self.finalize_withdrawals(&db, withdrawals)?;
+        self.finalize_deposits(&db, deposits)?;
+
```
jjyr, line 1336: In this line, when a new tip is synced, mem pool will try to collect deposits from CKB chain.
https://github.com/nervosnetwork/godwoken/blob/develop/crates/mem-pool/src/pool.rs#L816
We should change this line to:

```
if node_mode != NodeMode::ReadOnly {
  ... original logic
} else {
  self.pending_deposits.clear() // clear pending deposits
}
```


And this line, after synced new block, mem pool will try to write txs & withdrawals to current state.
https://github.com/nervosnetwork/godwoken/blob/develop/crates/mem-pool/src/pool.rs#L861

```
if node_mode != NodeMode::ReadOnly {
  self.prepare_next_mem_block(&db, withdrawals_iter, txs_iter)?;
}
```

jjyr (changes requested):

crates/mem-pool/src/sync/mq/gw_kafka.rs
```diff
@@ -0,0 +1,155 @@
...
+                    gw_types::packed::RefreshMemBlockMessageUnionReader::L2Transaction(tx) => {
+                        if let Err(err) = self.subscribe.add_tx(tx.to_entity()) {
+                            log::error!("[Subscribe tx] error: {:?}", err);
+                        }
+                    }
+                    gw_types::packed::RefreshMemBlockMessageUnionReader::NextMemBlock(next) => {
+                        match self.subscribe.next_mem_block(next.to_entity()) {
+                            Ok(None) => {
+                                log::debug!("Invalid tip. Wait for syncing to the new tip.");
+                                early_exit = Some(());
```
jjyr: why not use bool type?

zeroqn (dismissed):

crates/mem-pool/src/pool.rs
```diff
@@ -796,8 +815,11 @@ impl MemPool {
         }
 
         let db = self.store.begin_transaction();
-        // check pending deposits
-        self.refresh_deposit_cells(&db, new_tip)?;
+
+        if self.node_mode != NodeMode::ReadOnly {
+            // check pending deposits
+            self.refresh_deposit_cells(&db, new_tip)?;
```
jjyr, line 821: clear pending deposits if node_mode == ReadOnly

jjyr (approved):

zeroqn (approved):

PR #499, opened 2021-12-16
refactor: use snapshot to implement state db

jjyr:
Add a new structure MemStore:

``` rust
struct MemStore {
    snapshot: StoreSnapshot,
    mem: HashMap<_, _>
}
```

In this PR we share MemStore structure across modules which needs to access memory state instead of putting a kv map in the Store structure.

This PR also refactor current DB interface, we put mostly DB operations on traits: KVStoreRead & KVStoreWrite, and implement KVStoreRead for Store & StoreSnapshot. So we can read from snapshot instead of using DB transaction.


crates/rpc-server/src/registry.rs
```diff
@@ -366,7 +387,7 @@ impl RequestSubmitter {
             while let Some(hash) = mem_pool.pending_restored_tx_hashes().pop_front() {
                 match db.get_mem_pool_transaction(&hash) {
                     Ok(Some(tx)) => {
-                        if let Err(err) = mem_pool.push_transaction(tx) {
+                        if let Err(err) = smol::block_on(mem_pool.push_transaction(tx)) {
```
zeroqn, line 390: .await ?

crates/rpc-server/src/registry.rs
```diff
@@ -479,9 +501,9 @@ impl RequestSubmitter {
                 );
                 for entry in items {
                     let maybe_ok = match entry.item.clone() {
-                        FeeItem::Tx(tx) => mem_pool.push_transaction(tx),
+                        FeeItem::Tx(tx) => smol::block_on(mem_pool.push_transaction(tx)),
```
zeroqn, line 504: .await

crates/rpc-server/src/registry.rs
```diff
@@ -562,10 +584,10 @@ async fn get_transaction(
             (to_h256(tx_hash), verbose)
         }
     };
-    let db = store.begin_transaction();
+    let db = store.get_snapshot();
     let tx_opt;
     let status;
-    match db.get_transaction_info(&tx_hash)? {
+    match store.get_transaction_info(&tx_hash)? {
```
zeroqn: why not db.get_transaction_info? from snapshot?

crates/rpc-server/src/registry.rs
```diff
@@ -1084,20 +1111,22 @@ enum GetBalanceParams {
...
 ) -> Result<Uint128, RpcError> {
     let (short_address, sudt_id, block_number) = match params {
         GetBalanceParams::Tip(p) => (p.0, p.1, None),
         GetBalanceParams::Number(p) => p,
     };
 
-    let db = store.begin_transaction();
     let balance = match block_number {
         Some(block_number) => {
+            let db = store.begin_transaction();
```
zeroqn, line 1123: snapshot?

crates/rpc-server/src/registry.rs
jjyr:
```
    = help: within `MemStateTree<'_>`, the trait `Sync` is not implemented for `RefCell<HashSet<gw_common::H256>>`
    = note: required because it appears within the type `std::option::Option<RefCell<HashSet<gw_common::H256>>>`
    = note: required because it appears within the type `gw_store::state::state_tracker::StateTracker`
    = note: required because it appears within the type `MemStateTree<'_>`
    = note: required because of the requirements on the impl of `std::marker::Send` for `&MemStateTree<'_>`
```

If we call `.await` we need to implement Send on the MemStateTree, same for the other comment.
IMO we can figure out this solution in another PR.

zeroqn: Ok

zeroqn (approved):

magicalne (approved):

PR #505, opened 2021-12-20
Feat batch unlock withdrawal to owner

zeroqn:


crates/rpc-client/src/withdrawal.rs
```diff
@@ -0,0 +1,265 @@
...
+    last_finalized_block_number: u64,
+    l1_sudt_script_hash: &Byte32,
+) -> Result<()> {
+    verify_l1_sudt_script(info, l1_sudt_script_hash)?;
+    verify_finalized_owner_lock(info, last_finalized_block_number)
+}
+
+fn verify_l1_sudt_script(info: &CellInfo, l1_sudt_script_hash: &Byte32) -> Result<()> {
+    if let Some(sudt_type) = info.output.type_().to_opt() {
+        if info.data.len() < 16 {
```
Flouse: if info.data.len() < ckb_types::packed::Uint128::TOTAL_SIZE {

Flouse (dismissed):

Flouse (dismissed):

crates/block-producer/src/runner.rs
```diff
@@ -638,6 +651,28 @@ pub fn run(config: Config, skip_config_check: bool) -> Result<()> {
...
 
+            let unlocker_wallet = match block_producer_config
+                .withdrawal_unlocker_wallet_config
+                .as_ref()
+            {
+                Some(wallet_config) => {
+                    Wallet::from_config(wallet_config).with_context(|| "init unlocker wallet")?
+                }
+                None => {
+                    log::warn!("[unlock withdrawal] reuse block producer wallet");
```
jjyr: What is the benefit to use another wallet?

crates/block-producer/src/withdrawal.rs
```diff
@@ -180,3 +193,457 @@ pub fn revert(
...
+            &l1_sudt_script_hash,
+        ) {
+            log::error!("[unlock withdrawal] unexpected verify failed {}", err);
+            continue;
+        }
+
+        let owner_lock = {
+            let args: Bytes = withdrawal_cell.output.lock().args().unpack();
+            let lock_args_end = 32 + WithdrawalLockArgs::TOTAL_SIZE;
+            let lock_start = lock_args_end + 4;
```
jjyr: Why offset 4 bytes? We should add a comment to explain.

crates/block-producer/src/withdrawal.rs
```diff
@@ -180,3 +193,457 @@ pub fn revert(
...
+            let owner_lock = Script::new_unchecked(args.slice(lock_start..));
```
jjyr: slice will panic if lock_start is >= args.len(), we should check the length first?

crates/block-producer/src/withdrawal_unlocker.rs
```diff
@@ -0,0 +1,241 @@
...
+use crate::utils;
+
+pub use gw_rpc_client::contract::Guard;
+
+const TRANSACTION_FAILED_TO_RESOLVE_ERROR: &str = "TransactionFailedToResolve";
+
+pub struct FinalizedWithdrawalUnlocker {
+    unlocker: DefaultUnlocker,
+    unlocked_set: HashSet<OutPoint>,
+    unlock_txs: HashMap<H256, Vec<OutPoint>>,
```
jjyr, line 30: Is this a potential memory leak if the chain get reorged? Should we use lru structures to avoid it?

crates/generator/src/generator.rs
```diff
@@ -776,8 +791,9 @@ impl Generator {
         req: &WithdrawalRequest,
         block_hash: &H256,
         block_number: u64,
-        asset_script: Option<Script>,
-    ) -> Result<(CellOutput, Bytes), u128> {
+        opt_asset_script: Option<Script>,
+        opt_owner_lock: Option<Script>,
```
jjyr: We can use a enumerator struct to represent the two versions?

``` rust
enum UnlockWithdrawal {
  WithoutOwnerLock,
  WithOwnerLock {
    lock: Script
  }
}
```

crates/rpc-client/src/withdrawal.rs
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    Ok(())
+}
+
+fn verify_finalized_owner_lock(info: &CellInfo, last_finalized_block_number: u64) -> Result<()> {
+    let args: Bytes = info.output.lock().args().unpack();
+
+    let lock_args_end = 32 + WithdrawalLockArgs::TOTAL_SIZE;
+    let owner_lock_start = lock_args_end + 4;
```
jjyr: why offset by 4 bytes? need comment

crates/block-producer/src/withdrawal.rs
jjyr: We should put the parsing into a function.

zeroqn: Actually, ```args``` here is already checked by ```gw_rpc_client::withdrawal::verify_unlockable_to_owner``` fn.

crates/block-producer/src/withdrawal_unlocker.rs
zeroqn: We check ```unlock txs``` at https://github.com/nervosnetwork/godwoken/pull/505/files#diff-351c85653fef78eca92c9c1f32fc074f7ee69c132c2918799334c1dd6bd4d5f0R83.
If chain reorg happend, we will remove txs and out points.

jjyr (dismissed):

crates/rpc-client/src/withdrawal.rs
```diff
@@ -0,0 +1,265 @@
...
+        {
+            bail!("invalid l1 sudt script");
+        }
+    }
+
+    Ok(())
+}
+
+fn verify_finalized_owner_lock(info: &CellInfo, last_finalized_block_number: u64) -> Result<()> {
+    let args: Bytes = info.output.lock().args().unpack();
```
magicalne, line 37: All good, except this.
I think we should keep this implementation uniform. BTW this is a very black box.
How about extracting the same logic into crates/utils/src/withdrawal.rs. And verify there?

zeroqn: ```gw-utils``` current deps on ```gw-rpc-client``` and ```gw-poa```. We cannot extract this logic into ```gw-utils``` and use it here.
This will result in cyclic package dependency. We can refactor code in the future pr.
Right now, let tests cover our back.

magicalne (approved):

jjyr (approved):

PR #514, opened 2021-12-24
feat: collect error log and panic with sentry

magicalne:


crates/block-producer/src/runner.rs
```diff
@@ -400,6 +400,17 @@ impl BaseInitComponents {
...
+    // Set up sentry.
+    let _guard = match &config.sentry_dsn.as_ref() {
+        Some(sentry_dsn) => sentry::init((
+            sentry_dsn.as_str(),
+            sentry::ClientOptions {
+                release: sentry::release_name!(),
+                ..Default::default()
+            },
+        )),
+        None => sentry::init(()),
```
zeroqn, line 412: IMO, ```None``` is better then ```sentry::init(())```.

crates/block-producer/src/main.rs
```diff
@@ -127,6 +128,20 @@ fn run_cli() -> Result<()> {
...
     run_cli().expect("run cli");
 }
+
+fn init_log() {
+    let logger = env_logger::builder()
+        .parse_env(env_logger::Env::default().default_filter_or("info"))
+        .build();
+    let level = logger.filter();
+    let logger = sentry_log::SentryLogger::with_dest(logger).filter(|md| match md.level() {
+        log::Level::Error => LogFilter::Event,
```
jjyr: Should we also logging warning to sentry?

crates/block-producer/src/runner.rs
magicalne: Actually we used sentry-log as a wrapper here. I tried None instead of sentry::init() before. It turned out that the sentry had to be initialized here even if we don't enable sentry at all. Because env_logger is wrapped into it and we have to init env_logger by init sentry log. If we set None here, then we cannot get any printed log.

crates/block-producer/src/main.rs
magicalne: Let me try it.

magicalne: Just added. We don't have a lot of warning logs. Looks fine.

jjyr (approved):

zeroqn (approved):

PR #515, opened 2021-12-27
feat(block-producer): query contracts cell dep on chain

zeroqn:


crates/rpc-client/src/contract.rs
```diff
@@ -0,0 +1,163 @@
...
+        .build();
+    let type_script_hash = ckb_fixed_hash::H256(type_script_hash.into()).pack();
+
+    let search_key = SearchKey {
+        script: contract_type_script.into(),
+        script_type: ScriptType::Type,
+        filter: None,
+    };
+    let order = Order::Desc;
+    let limit = Uint32::from(DEFAULT_QUERY_LIMIT as u32);
```
jjyr: we only need to query 1 cell

jjyr (changes requested):

jjyr:
We should perform the searching before produce a block, so if the contract is updated, we could automatically update the reference.

zeroqn:
> We should perform the searching before produce a block, so if the contract is updated, we could automatically update the reference.

Done

crates/block-producer/src/runner.rs
```diff
@@ -232,9 +236,12 @@ pub struct BaseInitComponents {
     pub rpc_client: RPCClient,
     pub store: Store,
     pub generator: Arc<Generator>,
+    pub opt_block_producer_config: Option<BlockProducerConfig>,
+    pub contracts_dep_manager: Option<ContractsCellDepManager>,
 }
 
 impl BaseInitComponents {
+    #[allow(deprecated)]
```
jjyr, line 243: which function is deprecated?

crates/block-producer/src/runner.rs
```diff
@@ -433,17 +480,19 @@ pub fn run(config: Config, skip_config_check: bool) -> Result<()> {
     );
 
     let base = BaseInitComponents::init(&config, skip_config_check)?;
+    config.block_producer = base.opt_block_producer_config.clone();
```
jjyr: Why?, the config should be a immutable structure

crates/config/src/config.rs
```diff
@@ -109,26 +109,81 @@ pub struct ChallengerConfig {
...
+    #[serde(default)]
     pub allowed_eoa_deps: HashMap<H256, CellDep>,
+    #[deprecated]
+    #[serde(skip_serializing)]
+    #[serde(default)]
     pub allowed_contract_deps: HashMap<H256, CellDep>,
     pub challenger_config: ChallengerConfig,
     pub wallet_config: WalletConfig,
+    #[serde(default)]
+    pub contract_type_scripts: ContractTypeScriptConfig,
```
jjyr: This part is belongs to consensus, we I think maybe we can move it out from BlockProducer, and add a new consensus config?

``` rust
pub struct ConsensusConfig {
  pub contract_type_scripts: ContractTypeScriptConfig,
}
```

crates/block-producer/src/runner.rs
zeroqn: ```fn query_type_script_from_old_config()```

jjyr (dismissed):

crates/rpc-client/src/contract.rs
```diff
@@ -0,0 +1,250 @@
...
+impl ContractsCellDepManager {
+    pub async fn build(rpc_client: RPCClient, scripts: ContractTypeScriptConfig) -> Result<Self> {
+        let now = Instant::now();
+        let deps = query_cell_deps(&rpc_client, &scripts).await?;
+        log::trace!("[contracts dep] build {}ms", now.elapsed().as_millis());
+
+        Ok(Self {
+            rpc_client,
+            scripts: Arc::new(scripts),
+            deps: Arc::new(ArcSwap::new(Arc::new(deps))),
```
magicalne: This line makes me think for 20min at least... I suspect both "Arc"s are not necessary. 
I guess we can just use a normal pointer here guarded by ArcSwap.
```rust
deps: ArcSwap::from_pointee(deps);
```

zeroqn: Done

crates/rpc-client/src/contract.rs
```diff
@@ -0,0 +1,250 @@
...
+use crate::indexer_types::{Cell, Order, Pagination, ScriptType, SearchKey};
+use crate::rpc_client::RPCClient;
+use crate::utils::to_result;
+
+// Used in block producer and challenge
+#[derive(Clone)]
+pub struct ContractsCellDepManager {
+    rpc_client: RPCClient,
+    scripts: Arc<ContractTypeScriptConfig>,
+    deps: Arc<ArcSwap<ContractsCellDep>>,
```
magicalne, line 23: This Arc can also get removed. Elaborately **load()** and **store()** get auto deref twice here. I guess ArcSwap guards the pointee with AtoimicPointer with Strategy internally. So Arc takes no effect here?

```suggestion
    deps:ArcSwap<ContractsCellDep>,
```
This should be fine.

crates/rpc-client/src/contract.rs
```diff
@@ -0,0 +1,250 @@
...
+impl ContractsCellDepManager {
+    pub async fn build(rpc_client: RPCClient, scripts: ContractTypeScriptConfig) -> Result<Self> {
+        let now = Instant::now();
+        let deps = query_cell_deps(&rpc_client, &scripts).await?;
+        log::trace!("[contracts dep] build {}ms", now.elapsed().as_millis());
+
+        Ok(Self {
+            rpc_client,
+            scripts: Arc::new(scripts),
+            deps: Arc::new(ArcSwap::from_pointee(deps)),
```
magicalne, line 35:
```suggestion
            deps: ArcSwap::from_pointee(deps),
```

crates/rpc-client/src/contract.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    pub async fn refresh(&self) -> Result<()> {
+        log::info!("[contracts dep] refresh");
+
+        let now = Instant::now();
+        let deps = query_cell_deps(&self.rpc_client, &self.scripts).await?;
+        log::trace!("[contracts dep] refresh {}ms", now.elapsed().as_millis());
+
+        self.deps.store(Arc::new(deps));
```
magicalne, line 50:
```suggestion
self.deps.store(deps);
```

crates/rpc-client/src/contract.rs
zeroqn: No. ArcSwap doesn't implement Clone.

crates/rpc-client/src/contract.rs
zeroqn: Compile failure. Different type.

magicalne: You're right. This is fine.

magicalne (dismissed):

jjyr (dismissed):
Pls rebase

jjyr (approved):

PR #522, opened 2021-12-30
Fix AOT VM mode in generator

Flouse:
- fix `get_aot_code`
- fix: precompile AotCodes for VM0 and VM1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2211,7 +2211,7 @@ dependencies = [
 
 [[package]]
 name = "gw-generator"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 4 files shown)


.github/workflows/package.yaml
```diff
@@ -0,0 +1,13 @@
...
+on:
+  push:
+    branches:
+      - 'pkg/*'
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: full
+
+jobs:
```
jjyr: Jobs is empty?

Flouse: It was a mistake.
The unready `workflow` was removed.

jjyr (approved):

PR #529, opened 2022-01-03
feat(db): add db version and migrate

magicalne:
This pr includes:
- Check db version when we open rocksdb for the first time.
- Add a migrate trait that is used to migrate from an old version to the current one.
- It needs to init a db version as the first version.

### DIFF
--- a/crates/db/src/migrate.rs
+++ b/crates/db/src/migrate.rs
@@ -1,0 +1,223 @@
+// This mod is used to init a db version when the db version is absent at the first time.
+// And check present db version is still compatible. Godwoken must run on a valid db.
+// If godwoken with an advanced verion runs on an old db, this is the time we can run migrations.
+use crate::{
+    error::Error,
+    read_only_db::{self, ReadOnlyDB},
+    schema::{COLUMN_META, META_TIP_BLOCK_HASH_KEY},
+    Result,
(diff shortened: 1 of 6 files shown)


jjyr (dismissed):

crates/db/src/migrate.rs
```diff
@@ -0,0 +1,223 @@
...
+                        .to_string(),
+                })
+            }
+            Ordering::Equal => Ok(RocksDB::open(config, COLUMNS)),
+            Ordering::Less => {
+                log::info!("process fast migrations ...");
+                // TODO: Currently, there is only one migration: Add db version.
+                // So we can add db version by running default migration here.
+                // We should stop the process if we have more migrations and use **migration** command to run migration instead.
+                let factory = init_migration_factory();
```
Flouse: `factory` has been defined in L22.

Flouse (dismissed):

crates/db/src/migrate.rs
jjyr: @magicalne should remove this line?

magicalne: Definitely!

Flouse (approved):

zeroqn (approved):

PR #533, opened 2022-01-05
feat(config): load new configs dynamically through rpc

magicalne:


crates/rpc-server/src/registry.rs
```diff
@@ -441,7 +444,9 @@ impl RequestSubmitter {
                 let state = snap.state().expect("get mem state");
                 let kind = req.kind();
                 let hash = req.hash();
-                match self.req_to_entry(req, &state, queue.len()) {
+                let dynamic_config_manager = &self.dynamic_config_manager.read().await;
+                let fee_config = dynamic_config_manager.get_fee_config();
+                match req_to_entry(fee_config, self.generator.clone(), req, &state, queue.len()) {
```
Flouse, line 449: In this background loop,
we should consider the time comsumed by `generator.clone()`.

crates/rpc-server/src/registry.rs
```diff
@@ -457,7 +462,9 @@ impl RequestSubmitter {
                 while let Ok(req) = self.submit_rx.try_recv() {
                     let kind = req.kind();
                     let hash = req.hash();
-                    match self.req_to_entry(req, &state, queue.len()) {
+
+                    match req_to_entry(fee_config, self.generator.clone(), req, &state, queue.len())
```
Flouse: This operation is in the high frequency loop.

crates/dynamic-config/src/manager.rs
```diff
@@ -0,0 +1,72 @@
...
+// So that we don't need to restart to take effect every time.
+#[derive(Default)]
+pub struct DynamicConfigManager {
+    config_github_url: Option<String>,
+    fee_manager: FeeConfigManager,
+    whitelist_manager: WhilteListConfigManager,
+}
+
+impl DynamicConfigManager {
+    pub fn create(config: Config) -> Self {
```
jjyr, line 24: We should create new structure DynamicConfig or `RemoteConfig` to only include the configurations that can can be dynamically reloaded.

Use Config structure introcing too much complexity.

jjyr (changes requested):

crates/rpc-server/src/registry.rs
magicalne: That should be fine. This clone is an Arc clone. And we only read backends here which is fast. We can consider cache backends in the future maybe.

jjyr (approved):

zeroqn (approved):

PR #534, opened 2022-01-06
perf(mem-pool): reduce output block lock time

zeroqn:


crates/block-producer/src/block_producer.rs
```diff
@@ -385,17 +385,19 @@ impl BlockProducer {
...
         let (opt_finalized_custodians, block_param) = {
+            let output_fut = {
+                let t = Instant::now();
+                log::debug!("[compose_next_block_submit_tx] acquire mem-pool",);
+                let mem_pool = self.mem_pool.lock().await;
+                log::debug!(
+                    "[compose_next_block_submit_tx] unlock mem-pool {}ms",
+                    t.elapsed().as_millis()
+                );
+                mem_pool.output_mem_block(&OutputParam::new(retry_count))
```
magicalne, line 399: Can I say that this pr is about wrapping the old logic of **output_mem_block** into a future?
Could you please elaborate on why this change reduces the lock time of the output block?

jjyr: `output_mem_block` produces a future. This PR moves mostly calculation job into this future, so we can release mem-pool early.

IMO we can move the packaging logic out of mem-pool, this can makes mem-pool logic more clear.

zeroqn: I refactor logic out of mem-pool.
```
            let tip_block_number = mem_block.block_info().number().unpack().saturating_sub(1);
            let (finalized_custodians, produce_block_param) =
                generate_produce_block_param(&self.store, mem_block, post_block_state)?;
            let finalized_custodians = {
                let last_finalized_block_number = {
                    let context = self.generator.rollup_context();
                    context.last_finalized_block_number(tip_block_number)
                };
                let query = query_mergeable_custodians(
                    &self.rpc_client,
                    finalized_custodians.unwrap_or_default(),
                    last_finalized_block_number,
                );
                query.await?.expect_any()
            };

```

magicalne (approved):

jjyr (approved):

PR #563, opened 2022-01-21
feat(sync_mem_pool): use tokio based kafka impl

magicalne:


zeroqn (approved):

zeroqn:
Remove gw_kafka?

magicalne:
I'm thinking about removing it from another pr when it's stable. Just in case we need to revert back.

jjyr (approved):

PR #567, opened 2022-01-25
feat: add tracing

zeroqn:


crates/config/src/config.rs
```diff
@@ -10,6 +10,12 @@ use std::{
     path::PathBuf,
 };
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
```
magicalne, line 13: Do we really need `Copy` here?

zeroqn: ```Copy``` for this simply enum is cheap, just like u8.

magicalne (approved):

jjyr (approved):

PR #588, opened 2022-02-18
fix(mem-pool): none block median time error

zeroqn:
If tip block isn't on current canonical chain, we should try parent
block hash.

### DIFF
--- a/crates/mem-pool/src/default_provider.rs
+++ b/crates/mem-pool/src/default_provider.rs
@@ -1,9 +1,9 @@
 use std::time::Duration;
 
-use anyhow::{anyhow, Result};
+use anyhow::{bail, Result};
 use async_trait::async_trait;
 use gw_rpc_client::rpc_client::RPCClient;
-use gw_store::Store;
+use gw_store::{traits::chain_store::ChainStore, Store};
 use gw_types::{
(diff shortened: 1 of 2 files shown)


magicalne (approved):

Flouse (approved):

crates/mem-pool/src/default_provider.rs
```diff
@@ -30,15 +30,42 @@ impl DefaultMemPoolProvider {
...
+        let tip_l1_block_hash = tip_l1_block_hash_number.block_hash().unpack();
+        if let Some(median_time) = rpc_client.get_block_median_time(tip_l1_block_hash).await? {
+            return Ok(median_time - ONE_SECOND);
+        }
+
+        // tip l1 block hash is not on the current canonical chain, try parent block hash
+        // NOTE: Header doesn't include block hash
+        let mut l1_block_number = tip_l1_block_hash_number.number().unpack() + 1;
+        loop {
+            l1_block_number = l1_block_number.saturating_sub(1);
```
jjyr, line 49: IMO we should always retry `get_tip` RPC if we can't find tip block by tip hash. CKB node supposed to be updated to the new tip after the first call.

If we try to return median time by parent hash, we may use an outdated timestamp.

PR #599, opened 2022-02-24
Tokio metrics

magicalne:
- upgrade rust version
- add tokio-metrics and set rustflags with "tokio_unstable"
- monitor tokio runtime and some key tasks
- fix ci


.cargo/config
```diff
@@ -0,0 +1,3 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
```
jjyr: We shouldn't turn it on default.

PR #627, opened 2022-03-24
v1 fix: overwrite mem block timestamp if CKB median time is less than tip block

jjyr:
- https://github.com/nervosnetwork/godwoken/pull/625


zeroqn (approved):

Flouse (approved):

Flouse (changes requested):
Please fix the CI error:
https://github.com/nervosnetwork/godwoken/runs/5670799787?check_suite_focus=true#step:7:386

PR #630, opened 2022-03-25
docs: update `Life of a Polyjuice Transaction`

Flouse:
# Rename pETH to pCKB

## Why
- Developers and users feel confusing between pETH and ETH. Some of them thought that pETH is the real ETH.
- By default, we use CKB as **pCKB** in Polyjuice. CKB has 8 decimals while Ether has 18 decimals.
   > Shannon is a fractional denomination of CKBytes. One CKByte is equal to 100,000,000 Shannons.

I think the name `pCKB` is more clearer.

### DIFF
--- a/docs/known_caveats_of_polyjuice.md
+++ b/docs/known_caveats_of_polyjuice.md
@@ -1,65 +1,30 @@
 # Known Caveats Of Polyjuice
 
-When designing and building polyjuice, we aim at the highest level of compatibility, meaning:
+When designing and building Polyjuice, we aim at the highest level of compatibility, meaning:
 
-* The EVM used in polyjuice shall be almost 100% compatible with the latest fork version of Ethereum;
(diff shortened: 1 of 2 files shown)


jjyr:
Pls use another PR to submit code changes.

IMO we can use layer2-CKB to reference to the layer2 assets. Yokaiswap and other layer2 Dapps still use CKB as the symbol, so if we introduce pCKB in this documentation it may confuse users.

Flouse:
> Pls use another PR to submit code changes.

other code changes were removed.

janx:
`pCKB` and `layer2-CKB` are similar semantically, with different prefixes. I think  `pCKB` is better because its prefix is shorter, and `layer2-` refers to a much broader concept than Godwoken.

No matter it's `pCKB` or `xyzCKB`, it should be considered code name not display name. The applications should feel free to display `CKB` as token name on UI wherever it's `pCKB` behind.

Another option is to follow forcebridge's [convention](https://github.com/nervosnetwork/force-bridge/blob/main/configs/all-bridged-tokens.json), use `CKB|ckb` as symbol (code name).

jjyr (approved):

Issue #616, opened 2022-03-17
fix(gw-tools generate-config): set `producer.account_id` to the real one


Flouse:
In the current version, block_producer.account_id is default to 0.
We should set `producer.account_id` to the real one related to [block_producer.wallet_config].

In Godwoken v1, block producer should be an ETH EOA.

Related PR
- https://github.com/RetricSu/godwoken-kicker/pull/188

keroro520:
- https://github.com/nervosnetwork/godwoken/pull/634

Flouse:
- #634 was just a temporary solution.

IMO, it is better to remove `account_id` in [block_producer] config.

And we could get the block_producer.account_id through this process:
```
producer_private_key -> eth_address -> account_script (ETH EOA account script) 
-> account_script_hash -> account_id
```

see `build_eth_l2_script` function:
https://github.com/nervosnetwork/godwoken-polyjuice/blob/compatibility-breaking-changes/polyjuice-tests/src/helper.rs#L663-L672

Flouse:
> IMO, it is better to remove `account_id` in [block_producer] config.

`account_id` has been removed from [block_producer] config.
> https://github.com/nervosnetwork/godwoken/commit/bc8884aa768e3662fc70ada03d69a09012af3c19#diff-7cabbf516f45eaeed2dfa75063ad1da979b020f190a3bf6cdd4adaf23fe6826eL159

PR #634, opened 2022-03-27
fix(gw-tools): Use the first Godwoken account (id = 2) as block_producer

Flouse:
In Godwoken v1, block producer should be an ETH EOA.
By default, Use the first Godwoken EOA (account_id = 2) as block_producer.

- https://github.com/nervosnetwork/godwoken-polyjuice/pull/130
   `block.coinbase` needs the real block_producer's account ID

## Refs
- https://github.com/nervosnetwork/godwoken/issues/616
- https://github.com/RetricSu/godwoken-kicker/pull/188

### DIFF
--- a/crates/tools/src/generate_config.rs
+++ b/crates/tools/src/generate_config.rs
@@ -68,7 +68,7 @@ pub async fn generate_node_config(args: GenerateNodeConfigArgs<'_>) -> Result<Co
         .into();
 
     // build configuration
-    let account_id = 0;
+    let account_id = 2; // Use the first Godwoken account (id = 2) as block_producer by default
     let node_wallet_info = get_wallet_info(privkey_path);
     let code_hash: [u8; 32] = {
         let mut hash = [0u8; 32];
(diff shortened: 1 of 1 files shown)


crates/tools/src/generate_config.rs
```diff
@@ -68,7 +68,7 @@ pub async fn generate_node_config(args: GenerateNodeConfigArgs<'_>) -> Result<Co
         .into();
 
     // build configuration
-    let account_id = 0;
+    let account_id = 2; // Use the first Godwoken account (id = 2) as block_producer by default
```
zeroqn: Suggest `Use first godwoken eth account as block producer`

Flouse:
```suggestion
    // In Godwoken v1, block producer should be an ETH EOA.
    let account_id = 2; // Use the first Godwoken EOA (account_id = 2) as block_producer by default
```

zeroqn (approved):

PR #642, opened 2022-04-02
feat: mem pool p2p sync

blckngm:
Mem-pool syncing with p2p networking.

### DIFF
--- a/crates/mem-pool/src/sync/p2p.rs
+++ b/crates/mem-pool/src/sync/p2p.rs
@@ -1,0 +1,492 @@
+use std::{
+    collections::{BTreeMap, HashSet},
+    fmt,
+    sync::Arc,
+    time::Duration,
+};
+
+use futures::StreamExt;
+use gw_common::H256;
+use gw_p2p_network::{FnSpawn, P2P_MEM_BLOCK_SYNC_PROTOCOL, P2P_MEM_BLOCK_SYNC_PROTOCOL_NAME};
+use gw_types::{
+    packed::{
+        P2PSyncMessage, P2PSyncMessageReader, P2PSyncMessageUnion, P2PSyncMessageVec,
+        P2PSyncRequest, P2PSyncRequestReader, P2PSyncResponse, P2PSyncResponseReader,
+        P2PSyncResponseUnion, RefreshMemBlockMessageUnion, TipSync, TryAgain,
+    },
+    prelude::{Builder, Entity, Pack, Reader, Unpack},
+};
+use tentacle::{
+    builder::MetaBuilder,
+    error::SendErrorKind,
+    service::{ProtocolMeta, ServiceAsyncControl},
+    SessionId, SubstreamReadPart,
+};
+use tokio::sync::{broadcast, Mutex};
+
+use crate::pool::MemPool;
+
+const KEEP_BLOCKS: u64 = 3;
+
+#[derive(Default)]
+struct BlockMessages {
+    hash: H256,
+    messages: Vec<P2PSyncMessage>,
+}
+
+/// Buffer `P2PSyncMessage`s for recent blocks.
+#[derive(Default)]
+struct MessageBuffer {
+    // buffer[n] contains messages for block n.
+    buffer: BTreeMap<u64, BlockMessages>,
+}
+
+impl MessageBuffer {
+    fn push(&mut self, msg: P2PSyncMessage) {
+        if let Some((_, block)) = self.buffer.iter_mut().next_back() {
+            block.messages.push(msg);
+        }
+    }
+
+    fn handle_new_tip(&mut self, new_tip: (H256, u64), msg: &P2PSyncMessage) {
+        let later_blocks = self.buffer.split_off(&(new_tip.1 + 1));
+        self.buffer
+            .entry(new_tip.1)
+            .and_modify(|block| {
+                if block.hash != new_tip.0 || !later_blocks.is_empty() {
+                    // L1 reorg.
+                    block.hash = new_tip.0;
+                    // Clear messages.
+                    block.messages = vec![msg.clone()];
+                }
+            })
+            .or_insert_with(|| BlockMessages {
+                hash: new_tip.0,
+                messages: vec![msg.clone()],
+            });
+        self.buffer = self
+            .buffer
+            .split_off(&(new_tip.1.saturating_sub(KEEP_BLOCKS - 1)));
+    }
+
+    fn first_block_buffered(&self) -> Option<(H256, u64)> {
+        self.buffer.iter().next().map(|(k, v)| (v.hash, *k))
+    }
+
+    fn get_messages_after(&self, block: (H256, u64)) -> Option<P2PSyncMessageVec> {
+        if self.buffer.get(&block.1).map(|b| &b.hash) != Some(&block.0) {
+            return None;
+        }
+        let messages = self
+            .buffer
+            .range(block.1..)
+            .map(|(_, block)| block.messages.iter())
+            .flatten()
+            .cloned();
+        Some(P2PSyncMessageVec::new_builder().extend(messages).build())
+    }
+}
+
+#[derive(Default)]
+pub struct SyncServerState {
+    subscribers: HashSet<SessionId>,
+    buffer: MessageBuffer,
+}
+
+pub(crate) struct Publisher {
+    control: ServiceAsyncControl,
+    shared: Arc<Mutex<SyncServerState>>,
+}
+
+impl Publisher {
+    pub(crate) async fn handle_new_tip(&mut self, new_tip: (H256, u64)) {
+        let msg = P2PSyncMessage::new_builder()
+            .set(P2PSyncMessageUnion::TipSync(
+                TipSync::new_builder()
+                    .block_number(new_tip.1.pack())
+                    .block_hash(new_tip.0.pack())
+                    .build(),
+            ))
+            .build();
+        tracing::info!(tip = %HashAndNumber::from(new_tip), "publishing new tip");
+        let mut shared = self.shared.lock().await;
+        shared.buffer.handle_new_tip(new_tip, &msg);
+        for s in &shared.subscribers {
+            warn_result(
+                self.control
+                    .send_message_to(*s, P2P_MEM_BLOCK_SYNC_PROTOCOL, msg.as_bytes())
+                    .await,
+            );
+        }
+    }
+
+    pub(crate) async fn publish(&mut self, msg: RefreshMemBlockMessageUnion) {
+        let msg = match msg {
+            RefreshMemBlockMessageUnion::NextL2Transaction(tx) => {
+                let tx = tx.tx();
+                tracing::info!(hash = %hex::encode(&tx.hash()), "publishing L2Transaction");
+                P2PSyncMessageUnion::L2Transaction(tx)
+            }
+            RefreshMemBlockMessageUnion::NextMemBlock(b) => {
+                tracing::info!(
+                    number = b.block_info().number().unpack(),
+                    "publishing NextMemBlock"
+                );
+                P2PSyncMessageUnion::NextMemBlock(b)
+            }
+        };
+        let msg = P2PSyncMessage::new_builder().set(msg).build();
+        let mut shared = self.shared.lock().await;
+        shared.buffer.push(msg.clone());
+        for s in &shared.subscribers {
+            warn_result(
+                self.control
+                    .send_message_to(*s, P2P_MEM_BLOCK_SYNC_PROTOCOL, msg.as_bytes())
+                    .await,
+            );
+        }
+    }
+}
+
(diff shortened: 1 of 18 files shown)


magicalne (commented):
Not sure is there any way to drop old msgs in this pr? For example, a corner case could be: 
If the current tip of readonly node is 100, but the tip of the full node is 200. So we have to wait_on_tip for the readonly node to sync to the tip(200). After the readonly node sync to the new tip(200), the old msgs from the old tips should be removed.

blckngm:
> Not sure is there any way to drop old msgs in this pr? For example, a corner case could be: If the current tip of readonly node is 100, but the tip of the full node is 200. So we have to wait_on_tip for the readonly node to sync to the tip(200). After the readonly node sync to the new tip(200), the old msgs from the old tips should be removed.

When the readonly node gets to 200, it will send another `P2PSyncRequest { 200 }` requesting messages in and after 200.

zeroqn (dismissed):

magicalne (dismissed):

crates/p2p-network/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+
+#[async_trait]
+impl ServiceHandle for SHandle {
+    // A lot of internal error events will be output here, but not all errors need to close the service,
+    // some just tell users that they need to pay attention
+    async fn handle_error(&mut self, context: &mut ServiceContext, error: ServiceError) {
+        log::info!("service error: {:?}", error);
+        // Reconnect.
+        if let ServiceError::DialerError { address, error: _ } = error {
+            tokio::time::sleep(RECONNECT_DURATION).await;
```
driftluo: This is not recommended, this trait is on the main hub and may cause the entire server to be blocked， you can try to spawn a delay task to do this

crates/p2p-network/examples/p2p.rs
```diff
@@ -0,0 +1,54 @@
...
+use gw_config::P2PNetworkConfig;
+use gw_p2p_network::*;
+use tentacle::{builder::MetaBuilder, multiaddr::Protocol, service::ProtocolMeta, ProtocolId};
+
+const PROTOCOL_ZERO: ProtocolId = ProtocolId::new(0);
+
+fn protocol() -> ProtocolMeta {
+    MetaBuilder::new()
+        .id(PROTOCOL_ZERO)
+        .protocol_spawn(FnSpawn(|ctx, control, mut read| {
```
driftluo, line 14: The protocol name must be specified, it is the place where compatibility is guaranteed, the flag of the handshake, if not set, the default is "/p2p/protocol_id"

crates/p2p-network/examples/p2p.rs
```diff
@@ -0,0 +1,54 @@
...
+use futures_util::StreamExt;
+use gw_config::P2PNetworkConfig;
+use gw_p2p_network::*;
+use tentacle::{builder::MetaBuilder, multiaddr::Protocol, service::ProtocolMeta, ProtocolId};
+
+const PROTOCOL_ZERO: ProtocolId = ProtocolId::new(0);
+
+fn protocol() -> ProtocolMeta {
+    MetaBuilder::new()
+        .id(PROTOCOL_ZERO)
```
driftluo, line 13: https://github.com/nervosnetwork/axon/blob/main/core/network/src/protocols/mod.rs#L92

Consider using compression of messages to save bandwidth usage, with relatively high gains on large messages

crates/p2p-network/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+            let address: MultiAddr = d.parse().context("parse dial address")?;
+            dial.insert(address);
+        }
+        let dial_vec: Vec<MultiAddr> = dial.iter().cloned().collect();
+        let mut builder = ServiceBuilder::new()
+            .forever(true)
+            .tcp_config(|socket| {
+                let sock_ref = SockRef::from(&socket);
+                sock_ref.set_reuse_address(true)?;
+                sock_ref.set_nodelay(true)?;
```
driftluo, line 44: This setting has a great impact on the whole tcp, it is recommended to consider whether to turn on, no delay means that any small messages are sent, rather than having to send cache

crates/p2p-network/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        for d in &config.dial {
+            let address: MultiAddr = d.parse().context("parse dial address")?;
+            dial.insert(address);
+        }
+        let dial_vec: Vec<MultiAddr> = dial.iter().cloned().collect();
+        let mut builder = ServiceBuilder::new()
+            .forever(true)
+            .tcp_config(|socket| {
+                let sock_ref = SockRef::from(&socket);
+                sock_ref.set_reuse_address(true)?;
```
driftluo, line 43: Setting the reuse address but not the bind address is useless for dial operations

You can refer to https://github.com/nervosnetwork/axon/blob/main/core/network/src/service.rs#L266

blckngm: This is for the listening socket. Without this there will often be “address already in use” error on `bind` due to TIME_WAIT sockets in local (i.e. not docker) testing.

driftluo: This setting affects all incoming and outgoing sockets, not just the listening socket

crates/p2p-network/src/lib.rs
blckngm: Setting `TCP_NO_DELAY` disables Nagle's algorithm. It should not do much harm and may reduce latency when broadcasting.

crates/p2p-network/src/lib.rs
blckngm: Is there any harm for dial sockets?

Or how do I set it only for listening sockets?

crates/p2p-network/examples/p2p.rs
blckngm: OK. This is an example for testing. I will set names for the real protocols.

crates/p2p-network/src/lib.rs
blckngm: Noted.

crates/p2p-network/src/lib.rs
driftluo: If it's just for testing, consider differentiating the ports from the configuration.

> Is there any harm for dial sockets

It's just one more operation that causes other processes to use the same port, otherwise there is no effect for now

blckngm: Are you referring to `SO_REUSEPORT`? `SO_REUSEADDR` does not allow other processes bind to the exact same (address, port).

driftluo: https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ

These two parameters do not behave the same on different platforms, on windows it is possible, on linux you have to set both reuse port and reuse address

blckngm: Binding to the same port for outgoing sockets seems rather strange to me. Why do that?

Anyway, I don't think this has any harm for our outgoing sockets.

driftluo: I thought at first that you were planning to use the reuse port feature for public network autodiscovery, but if not, there's no harm in setting it up this way

crates/mem-pool/src/sync/p2p.rs
```diff
@@ -0,0 +1,465 @@
...
+                                .send_message_to(
+                                    context.id,
+                                    P2P_MEM_BLOCK_SYNC_PROTOCOL,
+                                    reply.as_bytes(),
+                                )
+                                .await,
+                        );
+                    }
+                }
+                while let Some(Ok(_)) = read_part.next().await {}
```
jjyr: Can you add a comment on this line?

crates/mem-pool/src/sync/p2p.rs
```diff
@@ -0,0 +1,465 @@
...
+            .map(|(_, block)| block.messages.iter())
+            .flatten()
+            .cloned();
+        Some(P2PSyncMessageVec::new_builder().extend(messages).build())
+    }
+}
+
+#[derive(Default)]
+pub struct SyncServerState {
+    subscribers: HashSet<SessionId>,
```
jjyr, line 92: How many subscribers can we support? Do we have a maximum limitation?

crates/mem-pool/src/sync/p2p.rs
```diff
@@ -0,0 +1,465 @@
...
+                tracing::info!(new_tip = %HashAndNumber::from(new_tip));
+                if new_tip == tip {
+                    return Ok(());
+                }
+                if new_tip.1 > tip.1 {
+                    return Err(new_tip);
+                }
+            }
+            Err(broadcast::error::RecvError::Lagged(_)) => {}
+            Err(_) => panic!("tip update sender closed"),
```
jjyr, line 270: I suggest return an error instead of panic. Does it panic on the program shutdown?

We can return an enum type error

``` rust
enum XX {
  UnexpectedTip {hash: H256, number: u64}
  Other(err)
}
```

crates/p2p-network/examples/p2p.rs
blckngm: Noted.

I think I will add compression (and authentication) when we allow other nodes to connect from the internet. Right now it's just for internal nodes.

crates/mem-pool/src/sync/p2p.rs
```diff
@@ -0,0 +1,465 @@
...
+        P2PSyncMessageUnion::TipSync(tip_sync) => {
+            let tip_sync: (H256, u64) = (
+                tip_sync.block_hash().unpack(),
+                tip_sync.block_number().unpack(),
+            );
+            tracing::info!(tip = %HashAndNumber::from(tip_sync), "handling tip sync");
+            // Wait for at least one update if block number goes back.
+            let at_least_once = tip_sync.1 < current_tip.1;
+            *current_tip = tip_sync;
+            let _ = wait_for_tip(mem_pool, *current_tip, at_least_once).await;
```
jjyr: We should print the error returned by this function

crates/p2p-network/examples/p2p.rs
blckngm: > compatibility is guaranteed

So how does this really work? How does the tentacle API support it? Say I want to upgrade a protocol but maintain backward compatibility with the origin protocol, is that what this is about?

driftluo: If the protocol name is the same, it will open with the latest common version.
If the corresponding protocol name is not found, then it will not be opened.

Current ckb supports 2 versions of protocol open at the same time:
https://github.com/nervosnetwork/ckb/blob/develop/network/src/protocols/support_protocols.rs#L98

When you open it, you can know the open version: https://github.com/nervosnetwork/ckb/blob/develop/network/src/protocols/discovery/mod.rs#L74

driftluo: There are two ways to maintain backward compatibility:
1. distinguish the behavior of the protocol with the version number and, after a period of parallelism, remove the support of the old version
2. reopen a new protocol and keep the old protocol support, and after a parallel period, turn off support for the old protocol

blckngm: OK. Got it.

crates/mem-pool/src/sync/p2p.rs
blckngm: This line is not reachable, because we are holding an reference to the mem_pool, which holds the sender, so the sender won't be closed. 

But we do need to consider how does this affect graceful shutdown.

crates/mem-pool/src/sync/p2p.rs
blckngm: Added

crates/mem-pool/src/sync/p2p.rs
blckngm: Added.

crates/mem-pool/src/sync/p2p.rs
blckngm: We may need to add authentication and whitelist for certain p2p features. Could be added later?

blckngm: I have added an issue for authentication and whitelist. Unresolve this if you think we need it (or to limit the number of subscribers) now.

crates/mem-pool/src/sync/p2p.rs
blckngm: I have added shutdown support for sync client protocol spawned tasks and tested that they work as expected.

(Sync server does not need any special handling because it only blocks on stream read so it ends naturally when we shutdown the whole p2p service and because it does not hold a reference to the mem pool (or any other important resources) so even if it does not shutdown it does not matter.)

jjyr (dismissed):

jjyr (dismissed):

blckngm:
Squashed and rebased.

jjyr (approved):

PR #644, opened 2022-04-02
feat(rpc): gw_get_mem_pool_state_ready

zeroqn:
Query whether a node is ready to serve

### DIFF
--- a/crates/store/src/mem_pool_state.rs
+++ b/crates/store/src/mem_pool_state.rs
@@ -1,4 +1,7 @@
-use std::sync::Arc;
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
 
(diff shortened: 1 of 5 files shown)


crates/store/src/mem_pool_state.rs
```diff
@@ -50,6 +55,14 @@ impl MemPoolState {
     pub fn store(&self, mem_store: Arc<MemStore>) {
         self.store.store(mem_store);
     }
+
+    pub fn complete_initial_syncing(&self) -> bool {
```
blckngm:
```suggestion
    pub fn completed_initial_syncing(&self) -> bool {
```

blckngm (changes requested):

blckngm (approved):

Flouse (approved):

PR #645, opened 2022-04-06
feat: fast withdrawal to v1

zeroqn:
- support generate and unlock withdrawal to v1 cell
- add withdrawal config

```rust
pub struct WithdrawalToV1Config {
    pub v1_rollup_type_hash: H256,
    pub v1_deposit_lock_code_hash: H256,
    pub v1_eth_lock_code_hash: H256,
    pub v1_deposit_minimal_cancel_timeout_msecs: u64,
}
```

- add `withdraw-to-v1` command to `gw-tools`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,7 +8,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 1
+        ref: ${{ github.event.pull_request.head.sha }}
     - name: Install Rust components
       run: rustup component add rustfmt && rustup component add clippy
     - uses: actions/cache@v2
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,9 +9,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-      with:
-        fetch-depth: 1
-        ref: ${{ github.event.pull_request.head.sha }}
     - name: Install Rust components
       run: rustup component add rustfmt && rustup component add clippy
     - uses: actions/cache@v2
--- a/crates/block-producer/src/withdrawal.rs
+++ b/crates/block-producer/src/withdrawal.rs
@@ -232,7 +232,7 @@ pub fn unlock_to_owner(
     let l1_sudt_script_hash = rollup_context.rollup_config.l1_sudt_script_type_hash();
     for withdrawal_cell in withdrawal_cells {
         // Double check
-        if let Err(err) = gw_rpc_client::withdrawal::verify_unlockable_to_owner(
+        if let Err(err) = gw_rpc_client::withdrawal::unlockable_to_owner(
             &withdrawal_cell,
             last_finalized_block_number,
             &l1_sudt_script_hash,
@@ -454,7 +454,7 @@ mod test {
         };
         let last_finalized_block_number =
             block.raw().number().unpack() + rollup_context.rollup_config.finality_blocks().unpack();
-        gw_rpc_client::withdrawal::verify_unlockable_to_owner(
+        gw_rpc_client::withdrawal::unlockable_to_owner(
             &info,
             last_finalized_block_number,
             &sudt_script.code_hash(),
--- a/crates/block-producer/src/withdrawal.rs
+++ b/crates/block-producer/src/withdrawal.rs
@@ -217,7 +217,7 @@ pub fn unlock_to_owner(
     let mut unlocked_to_owner_outputs = vec![];
 
     let unlock_via_finalize_witness = {
-        let unlock_args = UnlockWithdrawalViaFinalize::new_builder().build();
+        let unlock_args = UnlockWithdrawalViaFinalize::default();
         let unlock_witness = UnlockWithdrawalWitness::new_builder()
             .set(UnlockWithdrawalWitnessUnion::UnlockWithdrawalViaFinalize(
                 unlock_args,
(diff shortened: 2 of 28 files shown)


crates/block-producer/src/withdrawal.rs
```diff
@@ -226,38 +227,42 @@ pub fn unlock_to_owner(
             .lock(Some(unlock_witness.as_bytes()).pack())
             .build()
     };
+    let unlock_to_v1_witness = {
+        let unlock_args = UnlockWithdrawalToV1::new_builder().build();
```
blckngm:
```suggestion
        let unlock_args = UnlockWithdrawalToV1::default();
```

crates/rpc-client/src/withdrawal.rs
```diff
@@ -33,41 +56,85 @@ fn verify_l1_sudt_script(info: &CellInfo, l1_sudt_script_hash: &Byte32) -> Resul
...
+
+        let lock_args =
+            match WithdrawalLockArgsReader::verify(&args.slice(32..lock_args_end), false) {
+                Ok(()) => WithdrawalLockArgs::new_unchecked(args.slice(32..lock_args_end)),
+                Err(_) => bail!("invalid withdrawal lock args"),
+            };
+
+        let mut owner_lock_len_buf = [0u8; 4];
+        owner_lock_len_buf.copy_from_slice(&args.slice(lock_args_end..owner_lock_start));
+        let owner_lock_len = u32::from_be_bytes(owner_lock_len_buf) as usize;
```
blckngm: `u32::from_be_bytes(args[lock_args_end..owner_lock_start].try_into().unwrap())`?

(`args[lock_args_end..owner_lock_start]` COULD be a bit faster than `slice` because it does not need to create a new `Bytes` instance.)

Might also worth looking at `Buf` methods and see whether using them would make the code more clear, i.e. `get_u32`, `get_bytes`, etc: https://docs.rs/bytes/latest/bytes/buf/trait.Buf.html

.github/workflows/rust.yml
```diff
@@ -8,7 +8,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 1
+        ref: ${{ github.event.pull_request.head.sha }}
```
Flouse: Maybe we should rebase v0.10 and then run the test again.
Seems like this PR is conflicting with https://github.com/nervosnetwork/godwoken/commit/9a0602b47668cf5658df03f46aa75298081d2728.

Flouse (changes requested):

.github/workflows/rust.yml
zeroqn: ```acionts/checkout@v1``` somehow try to merge this pr with develop branch, not 0.10 branch.

I upgrade checkout to v3.

Flouse: This branch merged c33e49fe59b79e7d0877744dcd309dc3a07fc14a (feat-fast-withdrawal-to-v1) into e597b47bbe3cad9251a9b4c53925c3600ffa42df (v0.10).
Its [test](https://github.com/Flouse/godwoken/runs/6172402897?check_suite_focus=true) was failed.


```js
commit 4252c056887cf35a7f5f862d76a8e7244de1ee06 (HEAD)
Merge: e597b47b c33e49fe
Author: zeroqn <zero.forest@outlook.com>
Date:   Wed Apr 20 12:55:49 2022 +0000

    Merge c33e49fe59b79e7d0877744dcd309dc3a07fc14a (feat-fast-withdrawal-to-v1) into e597b47bbe3cad9251a9b4c53925c3600ffa42df (v0.10)
```

crates/block-producer/src/withdrawal.rs
```diff
@@ -226,38 +227,42 @@ pub fn unlock_to_owner(
             .lock(Some(unlock_witness.as_bytes()).pack())
             .build()
     };
+    let unlock_to_v1_witness = {
```
jjyr, line 230: We can implement this variable as a function variable to avoid runtime cost.

For example:

``` rust
let unlock_to_v1_witness = || {
  //..
};
```

.github/workflows/rust.yml
zeroqn: Please look at the error message
```
   --> crates/tools/src/withdraw_to_v1.rs:166:27
    |
166 |     let withdrawal_hash = godwoken_rpc_client.submit_withdrawal_request(bytes)?;
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the `?` operator cannot be applied to type `impl std::future::Future`
```
It complain about ```Future```, but `godwoken_rpc_client.submit_withdrawal_request` in v0 isn't ```async fn```.
It's ```async fn``` in v1.

Please clone the pr and run test local.

zeroqn: My mistake.
Didn't notice new commits...
I'll rebase code.

crates/tools/src/main.rs
```diff
@@ -348,6 +349,55 @@ async fn run_cli() -> Result<()> {
...
+                        .required(true)
+                        .help("v1 eth address"),
+                )
+                .arg(
+                    Arg::with_name("sudt-script-hash")
+                        .long("sudt-script-hash")
+                        .takes_value(true)
+                        .required(false)
+                        .default_value(
+                            "0x0000000000000000000000000000000000000000000000000000000000000000",
```
jjyr, line 396: Is this default value valid? I suggest we remove the default value if it is invalid.

jjyr (changes requested):

crates/tools/src/main.rs
zeroqn: Its same as "withdraw" command. 

https://github.com/nervosnetwork/godwoken/blob/43b7427b3a40d85ee439285865863fec5a057648/crates/tools/src/main.rs#L347

crates/block-producer/src/withdrawal.rs
jjyr: We should apply the same trick on `unlock_via_finalize_witness` variable @zeroqn

jjyr (approved):

blckngm (approved):

Flouse (approved):

PR #654, opened 2022-04-09
V1: feat(gw-tools): add command report-accounts to export account infos

jjyr:
- https://github.com/nervosnetwork/godwoken/pull/653

### DIFF
--- a/crates/tools/src/main.rs
+++ b/crates/tools/src/main.rs
@@ -13,6 +13,7 @@ pub mod godwoken_rpc;
 mod hasher;
 mod polyjuice;
 mod prepare_scripts;
+mod report_accounts;
 mod setup;
 mod stat;
 mod sudt;
@@ -788,6 +789,19 @@ async fn run_cli() -> Result<()> {
                 ),
         )
         .subcommand(
+            SubCommand::with_name("report-accounts")
+                .about("Report account to csv files")
+                .arg(arg_godwoken_rpc_url.clone())
+                .arg(
+                    Arg::with_name("output")
+                        .short("o")
+                        .long("output")
+                        .takes_value(true)
+                        .required(true)
+                        .help("output file"),
+                ),
+        )
+        .subcommand(
             SubCommand::with_name("stat-custodian-ckb")
                 .about("Output amount of layer2 custodian CKB")
                 .arg(arg_indexer_rpc.clone())
@@ -1026,7 +1040,9 @@ async fn run_cli() -> Result<()> {
                 ckb_rpc_url.as_str(),
                 eth_address,
                 godwoken_rpc_url,
-            ) {
+            )
+            .await
+            {
                 log::error!("Deposit CKB error: {}", err);
                 std::process::exit(-1);
             };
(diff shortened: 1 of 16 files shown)


crates/tools/src/main.rs
```diff
@@ -787,6 +788,19 @@ async fn run_cli() -> Result<()> {
                         .help("output file"),
                 ),
         )
+        .subcommand(
+            SubCommand::with_name("report-accounts")
+                .about("Report account to csv files")
```
blckngm, line 793:
```suggestion
                .about("Report accounts to csv file")
```

crates/tools/src/report_accounts.rs
```diff
@@ -0,0 +1,72 @@
...
+use std::path::Path;
+use tokio::task::JoinHandle;
+
+use crate::godwoken_rpc::GodwokenRpcClient;
+
+#[derive(Serialize, Debug)]
+struct Account {
+    pub id: u32,
+    pub code_hash: ckb_fixed_hash::H256,
+    pub ckb: u128,
```
blckngm, line 13: What tools do we intend to interoperate with with the csv format? Do they support `u128`? If not then it might make sense to include a string or floating point representation as well.

crates/tools/src/report_accounts.rs
```diff
@@ -0,0 +1,72 @@
...
+    let mut wtr = csv::Writer::from_path(output)?;
+    while let Some(task) = rx.recv().await {
+        match task.await?? {
+            Some(account) => {
+                wtr.serialize(account)?;
+                total_account += 1;
+                print!(".")
+            }
+            None => {
+                producer_handle.abort();
```
blckngm, line 64: Took me a while to understand the control flow.

There's a small caveat: the producer task is not aborted if some error happens. (It will still end when `rx` is dropped and `send` fails, and even if it does not it wouldn't really matter in a cli tool. But in long running services we should be more careful, e.g. by following the **structured concurrency** paradigm and use more reliable task joining/cancellation constructs.)

Also, could [`buffered`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffered) be used to simplify this code a bit?

blckngm: About stream buffering: https://gendignoux.com/blog/2021/04/01/rust-async-streams-futures-part1.html#ordered-buffering

crates/tools/src/report_accounts.rs
jjyr: You are right, but for now we only want a simple analysis tools, csv format is enought for now. We can support export data in other formats when we really need the feature.

blckngm: I was worrying about other csv parsers not able to handle large integers. But yeah, let's deal with it when specific problem arises.

crates/tools/src/report_accounts.rs
jjyr: Yes, if we use `buffered` we may remove the channel, but we also use the channel to control the concurrent jobs of the 'producer', I think it is reliable enough for the cli like you said. With `buffered` feature I don't know how to do this.

blckngm: The `n` parameter limits concurrency in `buffered`. No more than `n` futures will be polled concurrency and the result is produced in order. Usage is usually like this:

```rust
let stream = stream::iter(0..)
    .map(|x| async move {
        // async request with x
        x
    })
    .buffered(20)
    .take_while(...);
// Do what you want with the results in stream.
```

(Not saying that you have to use this, just providing an alternative approach.)

blckngm (approved):

PR #655, opened 2022-04-13
feat: get in queue requests

blckngm:
Now the `gw_get_transaction`/`gw_get_withdrawal` RPC calls will return

```
{ "status": "in_queue", ... }
```

instead of null for in queue transactions/withdrawals.

On a full node `gw_get_transaction`/`gw_get_withdrawal` will be almost
always accurate.

Readonly nodes when well connected to the full node via the p2p network
will be notified about in queue request changes, so they will also
return

```
{ "status": "in_queue", ... }
```

for requests that **they know** are in queue.

But due to network delay or connectivity issues they may still return
inaccurate answers, e.g. they may return null for some in queue
requests, or they may return “in queue” or null for actually
pending/committed requests.


jjyr:
This PR introduce too many cost to the protocol compares to the benefit. We should give a better solution.

Issue #507, opened 2021-12-20
Speed up node block synchronization time


e00dan:
For blocks full of transactions on testnet it can take more than 2 minutes to sync a block (longer then block produce time [~1 min?], which results in not being able to catch up with the tip):

eg. block 205520 took 2 mins 27 seconds to sync 

```
godwoken_1  | [2021-12-20T21:35:06Z INFO  gw_block_producer::poller] known last tx hash, skip 0ae1bfc7dada69ee4a822193eab79466849c33e4d47132eefa98355f35016d4d
godwoken_1  | [2021-12-20T21:36:03Z INFO  gw_chain::chain] sync new block #205502 success
godwoken_1  | [2021-12-20T21:36:20Z INFO  gw_block_producer::poller] known last tx hash, skip ca6714d068569357e64cd1090cd721ff518b7ca847dae60af0d31de72d2e9fd8
godwoken_1  | [2021-12-20T21:38:11Z INFO  gw_chain::chain] sync new block #205503 success
godwoken_1  | [2021-12-20T21:38:31Z INFO  gw_block_producer::poller] known last tx hash, skip 69135636c50a9685d018eafcfd483caaab30fa756113daa216934cf7e6d8ebf0
godwoken_1  | [2021-12-20T21:39:42Z INFO  gw_chain::chain] sync new block #205504 success
godwoken_1  | [2021-12-20T21:39:54Z INFO  gw_block_producer::poller] known last tx hash, skip a61b7b1ec49a659b113ecc4152ae5546c522cf7983c93f876f2a82fce6229166
godwoken_1  | [2021-12-20T21:41:34Z INFO  gw_chain::chain] sync new block #205505 success
godwoken_1  | [2021-12-20T21:41:51Z INFO  gw_block_producer::poller] known last tx hash, skip 318644f963d763b03bcb350792cf3fb99c054c33b53b924348628720e25d5767
godwoken_1  | [2021-12-20T21:43:04Z INFO  gw_chain::chain] sync new block #205506 success
godwoken_1  | [2021-12-20T21:43:21Z INFO  gw_block_producer::poller] known last tx hash, skip bcdfe7d11254fdc9c8d1fc261a034b0ca9d7aa0fc1dc26f9ce1346cb7d47a720
godwoken_1  | [2021-12-20T21:45:10Z INFO  gw_chain::chain] sync new block #205507 success
[...]

PR #665, opened 2022-04-21
perf: optimize syncing from L1

blckngm:
Optimize syncing from L1 by pipelining ckb/ckb-indexer requests and the actual syncing.

Generally, syncing should be a bit faster.

This is particularly helpful when ckb/ckb-indexer request-response time is bit high, cc #507.

(This also updates opentelemetry-jaeger and enables batch auto split to avoid message too large errors when exporting trace spans to jaeger.)


MrPiNkY2022 (commented):
Approve

Cargo.lock
```diff
@@ -3479,13 +3481,15 @@ dependencies = [
 
 [[package]]
 name = "opentelemetry"
-version = "0.16.0"
```
MrPiNkY2022, line 3482: +version = "0.17.0"

MrPiNkY2022 (approved):

jjyr:
Can you update this PR?

PR #675, opened 2022-04-25
BREAKING CHANGE: change ckb decimal to 18

zeroqn:
## Purpose

This PR unified layer2 fungible token representation as u256.

This PR also changes the decimal of CKB, in layer1 CKB is represented in 8 decimal, but when user deposit CKB into layer2, we changed the representation to 18 decimal. Thus, layer2 CKB can play the same role as ETH, developers can operate CKB as 18 decimal asset(like ETH) in their contract without modification. This change also improve the compatibility between Metamask and the native token CKB.


crates/common/src/ckb_decimal.rs
```diff
@@ -0,0 +1,34 @@
+use gw_types::U256;
+
+// NOTE: u64::MAX is about 10^18, U256::MAX is about 10^77. Multiple 10 pow
+// `CKB_DECIMAL_POW_EXP` should not overflow.
+pub const CKB_DECIMAL_POW_EXP: u32 = 10;
+pub const CKB_DECIMAL_POWER_TEN: u64 = 10u64.pow(CKB_DECIMAL_POW_EXP);
+
+pub fn to_18(amount: u64) -> U256 {
```
magicalne:
```suggestion
pub fn to_u256(amount: u64) -> U256 {
```

crates/common/src/ckb_decimal.rs
```diff
@@ -0,0 +1,34 @@
...
+    U256::from(amount) * CKB_DECIMAL_POWER_TEN
+}
+
+pub fn from_18(amount: U256) -> Option<u64> {
```
magicalne:
```suggestion
pub fn from_u256(amount: U256) -> Option<u64> {
```

crates/common/src/ckb_decimal.rs
zeroqn: I prefer keep decimal info in function name. Since every sudt amount is U256 now. ```ckb_decimal::to_18``` is better than ```ckb_decimal::to_u256```.

crates/common/src/ckb_decimal.rs
```diff
@@ -0,0 +1,34 @@
...
+    let truncated = amount / CKB_DECIMAL_POWER_TEN;
+    if truncated.bits() > u64::BITS as usize {
+        None
+    } else {
+        Some(truncated.as_u64())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
blckngm: It's good to be cautious, but overflows in constant expressions can be caught at compile time, so the test is a bit redundant.

crates/common/src/state.rs
```diff
@@ -229,20 +231,21 @@ pub trait State {
         &mut self,
         sudt_id: u32,
         address: &RegistryAddress,
-        amount: u128,
+        amount: impl Into<U256>,
```
blckngm: Generics like this if monomorphized many times can increase increase compile time and binary size.

I suggest that you just use `amount: U256` here and let the caller call `.into()`.

crates/common/src/state.rs
```diff
@@ -254,20 +257,21 @@ pub trait State {
...
```
blckngm:
```suggestion
        amount: U256,
```

crates/common/src/ckb_decimal.rs
```diff
@@ -0,0 +1,17 @@
...
+
+pub fn to_18(amount: u64) -> u128 {
+    u128::from(amount) * CKB_DECIMAL_POWER_TEN
+}
+
+pub fn from_18(amount: u128) -> Option<u64> {
+    amount
+        .checked_div(CKB_DECIMAL_POWER_TEN)
+        .map(|a| a.try_into().ok())
+        .flatten()
```
blckngm: `.map(f).flatten()` could be simplified a bit as `.and_then(f)`

jjyr:
We should also change `SUDTTransfer#amount` to Uint256, since the layer2 is using decimal 18. The burn / mint interface should also be updated.

crates/generator/src/traits.rs
```diff
@@ -171,10 +175,13 @@ impl<S: State + CodeStore> StateExt for S {
...
         };
-        self.mint_sudt(CKB_SUDT_ACCOUNT_ID, &address, capacity.into())?;
+        // Align CKB to 18 decimals
+        let ckb_amount = ckb_decimal::to_18(capacity);
+        self.mint_sudt(CKB_SUDT_ACCOUNT_ID, &address, ckb_amount)?;
         log::debug!(
-            "[generator] mint {} shannons CKB to account {}",
-            capacity,
+            "[generator] mint {} shannons * 10^{} CKB to account {}",
+            ckb_amount,
```
Flouse:
```suggestion
            capacity,
```

Flouse (changes requested):

Flouse (dismissed):

crates/generator/src/traits.rs
jjyr: @zeroqn Please correct this log message.

Flouse (dismissed):

Flouse (dismissed):

crates/common/src/ckb_decimal.rs
```diff
@@ -0,0 +1,14 @@
...
+// NOTE: u64::MAX is about 10^18, U256::MAX is about 10^77. Multiple 10 pow
+// `CKB_DECIMAL_POW_EXP` should not overflow.
+pub const CKB_DECIMAL_POW_EXP: u32 = 10;
+pub const CKB_DECIMAL_POWER_TEN: u64 = 10u64.pow(CKB_DECIMAL_POW_EXP);
+
+pub fn to_18(amount: u64) -> U256 {
+    U256::from(amount) * CKB_DECIMAL_POWER_TEN
+}
+
+pub fn from_18(amount: U256) -> U256 {
```
keroro520:
```suggestion
pub fn from_18(amount: U256) -> u64 {
```

Avoid maintaining two kinds of `U256` with different precisions.

keroro520 (changes requested):

jjyr (dismissed):

crates/common/src/ckb_decimal.rs
jjyr: Good point, I think we should using a wrapped type to avoid conversion by hand.

```
CKBCapacity(U256);

impl CKBCapacity {
  fn to_layer1() -> u64 {}
  fn to_layer2() -> U256 {}
  fn from_layer1(n: u64) -> Self {}
  fn from_layer2(n: U256) -> Self {}
}
```

This provides more clue on the source code than the current version:

```
CKBCapacity::from_layer1(deposit.capacity()).to_layer2()

// VS

to_18(deposit.capacity())
```

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ In the new version, we improve the compatibility of Godwoken:
 - Support Ethereum signature format and EIP-712. User can view the transaction before signing, instead of signing a random 32 bytes message. [#561](https://github.com/nervosnetwork/godwoken/pull/561)
 - Fix the total supply interface of sUDT ERC-20 proxy contract [#560](https://github.com/nervosnetwork/godwoken/pull/560)
 - Support interactive with eth address that haven't been registered to Godwoken.
+- Unify layer 2 fungible token represatation as uint256.
+- Change layer 2 ckb decimal from 8 to 18, improve compatibility between metamask and native ckb. [#675](https://github.com/nervosnetwork/godwoken/pull/675)
 
 In short, as a developer, you can use Godwoken v1 just like anyother Ethereum compatible chain, all you need to do is to switch the network to Godwoken. The polyjuice-provider web3 plugin is removed in the v1 version.
```
ashleewang888: In short, developers can use Godwoken v1 the same way as other ethereum compatible chains, all that has to be done is to switch the network to Godwoken. The polyjuice-provider web3 plugin has been removed in v1.

crates/mem-pool/src/types.rs
```diff
@@ -49,7 +50,8 @@ impl EntryList {
         // remove lower balance withdrawals
         if let Some(withdrawal) = self.withdrawals.get(0) {
             let withdrawal_capacity: u64 = withdrawal.raw().capacity().unpack();
-            if (withdrawal_capacity as u128) > capacity {
+            let capacity = capacity.to_layer1().unwrap_or(u64::MAX);
```
keroro520: What cases go into the `.unwrap_or(u64::MAX)`?

CHANGELOG.md
zeroqn: Yes

crates/mem-pool/src/types.rs
zeroqn: We can directly unwrap here. Just code style.

keroro520: I think `unwrap` directly is better.

keroro520 (dismissed):

crates/mem-pool/src/types.rs
zeroqn: Ok

Flouse (dismissed):

keroro520 (dismissed):

jjyr (approved):

Flouse (approved):

PR #626, opened 2022-03-23
perf: optimize molecule usage

blckngm:
The principle here is to try to avoid `Entity::from_slice` or `Reader::to_entity` if possible, because they need to allocate and copy memory.

* Replace `Box<[u8]>` -> `as_ref` -> `Reader::from_slice_should_be_ok` -> `to_entity` with `from_box_should_be_ok!`, which is `from_slice_should_be_ok` + `into` -> `new_unchecked`.

* Replace `Entity::from_slice` -> `as_reader` with `Reader::from_slice`.

* Replace `to_entity` -> `unpack` with `unpack`.

PR #677, opened 2022-04-26
Refactor tx verification

jjyr:
1. Move tx & withdrawal verification code to a standalone module.
2. Reject tx if sender's balance is insufficient.

### DIFF
--- a/crates/generator/src/typed_transaction/types.rs
+++ b/crates/generator/src/typed_transaction/types.rs
@@ -1,0 +1,144 @@
+use std::convert::TryInto;
+
+use ckb_vm::Bytes;
+use gw_common::builtins::CKB_SUDT_ACCOUNT_ID;
+use gw_types::{
+    core::AllowedContractType,
+    packed::{ETHAddrRegArgs, L2Transaction, MetaContractArgs, SUDTArgs},
+    prelude::*,
+};
+
+use crate::error::{AccountError, TransactionValidateError};
+
+/// Types Transaction
+pub enum TypedTransaction {
+    EthAddrReg(EthAddrRegTx),
+    Meta(MetaTx),
+    SimpleUDT(SimpleUDTTx),
+    Polyjuice(PolyjuiceTx),
+}
+
+impl TypedTransaction {
+    pub fn from_tx(
+        tx: L2Transaction,
+        type_: AllowedContractType,
+    ) -> Result<Self, TransactionValidateError> {
+        let tx = match type_ {
+            AllowedContractType::EthAddrReg => Self::EthAddrReg(EthAddrRegTx(tx)),
+            AllowedContractType::Meta => Self::Meta(MetaTx(tx)),
+            AllowedContractType::Sudt => Self::SimpleUDT(SimpleUDTTx(tx)),
+            AllowedContractType::Polyjuice => Self::Polyjuice(PolyjuiceTx(tx)),
+            AllowedContractType::Unknown => return Err(AccountError::UnknownScript.into()),
+        };
+        Ok(tx)
+    }
+
+    /// Got expect cost of the tx, (transfer value + fee).
+    /// returns none if tx has no cost, it may happend when we call readonly interface of some Godwoken builtin contracts.
+    pub fn cost(&self) -> Option<u64> {
+        match self {
+            Self::EthAddrReg(tx) => tx.cost(),
+            Self::Meta(tx) => tx.cost(),
+            Self::SimpleUDT(tx) => tx.cost(),
+            Self::Polyjuice(tx) => tx.cost(),
+        }
+    }
+}
+
+pub struct EthAddrRegTx(L2Transaction);
+
+impl EthAddrRegTx {
+    pub fn cost(&self) -> Option<u64> {
+        use gw_types::packed::ETHAddrRegArgsUnion::*;
+
+        let args: Bytes = self.0.raw().args().unpack();
+        let args = ETHAddrRegArgs::from_slice(&args).ok()?;
+
+        match args.to_enum() {
+            EthToGw(_) | GwToEth(_) => None,
+            SetMapping(args) => Some(args.fee().amount().unpack()),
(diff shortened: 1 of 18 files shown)


magicalne (dismissed):

crates/generator/src/typed_transaction/types.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+pub struct EthAddrRegTx(L2Transaction);
+
+impl EthAddrRegTx {
+    pub fn cost(&self) -> Option<u64> {
+        use gw_types::packed::ETHAddrRegArgsUnion::*;
+
+        let args: Bytes = self.0.raw().args().unpack();
+        let args = ETHAddrRegArgs::from_slice(&args).ok()?;
```
blckngm: Could `ETHAddrRegArgsReader` be used instead (to avoid allocation and copy)?

crates/tests/src/testing_tool/chain.rs
```diff
@@ -145,6 +145,21 @@ lazy_static! {
         hasher.finalize(&mut buf);
         buf
     };
+    pub static ref SUDT_VALIDATOR_PROGRAM: Bytes = {
+        let mut buf = Vec::new();
+        let mut path = PathBuf::new();
+        path.push(&SUDT_VALIDATOR_PATH);
+        let mut f = fs::File::open(&path).expect("load SUDT program");
+        f.read_to_end(&mut buf).expect("read SUDT program");
```
blckngm: Why not just use `fs::read(&SUDT_VALIDATOR_PATH).expect(...).into()`?

crates/generator/src/typed_transaction/types.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+pub struct MetaTx(L2Transaction);
+
+impl MetaTx {
+    pub fn cost(&self) -> Option<u64> {
+        use gw_types::packed::MetaContractArgsUnion::*;
+
+        let args: Bytes = self.0.raw().args().unpack();
+        let args = MetaContractArgs::from_slice(&args).ok()?;
```
blckngm: Use `MetaContractArgsReader` instead?

crates/generator/src/typed_transaction/types.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+pub struct SimpleUDTTx(L2Transaction);
+
+impl SimpleUDTTx {
+    pub fn cost(&self) -> Option<u64> {
+        use gw_types::packed::SUDTArgsUnion::*;
+
+        let args: Bytes = self.0.raw().args().unpack();
+        let args = SUDTArgs::from_slice(&args).ok()?;
```
blckngm: `SUDTArgsReader`?

crates/generator/src/typed_transaction/types.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+pub struct PolyjuiceTx(L2Transaction);
+impl PolyjuiceTx {
+    pub fn cost(&self) -> Option<u64> {
+        let args: Bytes = self.0.raw().args().unpack();
+        if args.len() < 52 {
+            log::error!(
+                "[gw-generator] parse PolyjuiceTx error, wrong args.len expected: 52, actual: {}",
```
blckngm:
```suggestion
                "[gw-generator] parse PolyjuiceTx error, wrong args.len expected: >= 52, actual: {}",
```

crates/generator/src/typed_transaction/types.rs
jjyr: molecule structures is implemented as 'zero-copy' under the 'std' rust environment, only the reference of a pointer is cloned.

I choose to use non reader version to avoid too many lifetime notations.

blckngm: My reading is that `Reader::from_slice` and `Entity::new_unchecked` are zero-copy, but `Entity::from_slice` or `Reader::to_entity` are not:

```rust
impl<'r> molecule::prelude::Reader<'r> for SUDTArgsReader<'r> {
    fn to_entity(&self) -> Self::Entity {
        Self::Entity::new_unchecked(self.as_slice().to_owned().into())
    }
}
```

Lifetime annotations are not necessary. See https://github.com/nervosnetwork/godwoken/pull/626/files#diff-737eb91a5e5aa0d1609b8bd48181fd1b9d647de85f773ad585172da442e654d7

crates/generator/src/verification/transaction.rs
```diff
@@ -0,0 +1,108 @@
...
+            .get_registry_address_by_script_hash(ETH_REGISTRY_ACCOUNT_ID, &sender_script_hash)?
+            .ok_or(AccountError::RegistryAddressNotFound)?;
+        let balance = self
+            .state
+            .get_sudt_balance(CKB_SUDT_ACCOUNT_ID, &sender_address)?;
+        // get balance
+        let tx_cost = {
+            let tx_type = self.get_tx_type(tx)?;
+            let typed_tx = TypedTransaction::from_tx(tx.to_owned(), tx_type)?;
+            typed_tx.cost().map(Into::into).unwrap_or(u128::MAX)
```
blckngm: The handling of `None` here does not seem to match the doc comment of the `cost` method:

```
     /// returns none if tx has no cost, it may happend when we call readonly interface of some Godwoken builtin contracts.
```

crates/generator/src/typed_transaction/types.rs
jjyr: You are right, I misunderstood.

crates/generator/src/verification/transaction.rs
jjyr: Right, I will add another comment on this line.

jjyr: Please review.

blckngm (approved):

zeroqn (approved):

magicalne (approved):

PR #682, opened 2022-05-09
fix(withdrawal_unlocker): rejected unlock tx isn't removed

zeroqn:
- fix ```get_transaction_status``` throw ```invalid type: null, expected struct TransactionView``` error for rejected tx, return  ```Ok(None)``` instead.
- always drop unlock tx and retry if ```get_transaction_status``` return error, avoid "lock" withdrawal cell.

### DIFF
--- a/crates/block-producer/src/withdrawal_unlocker.rs
+++ b/crates/block-producer/src/withdrawal_unlocker.rs
@@ -84,7 +84,9 @@ impl FinalizedWithdrawalUnlocker {
         for (tx_hash, withdrawal_to_unlock) in self.unlock_txs.iter() {
             match rpc_client.get_transaction_status(*tx_hash).await {
                 Err(err) => {
-                    log::debug!("[unlock withdrawal] get unlock tx failed {}", err);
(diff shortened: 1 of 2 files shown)


jjyr (approved):

blckngm:
> fix get_transaction_status throw invalid type: null, expected struct TransactionView error for rejected tx, return Ok(None) instead.

It seems that this could be fixed by updating ckb-jsonrpc-types too: https://docs.rs/ckb-jsonrpc-types/latest/ckb_jsonrpc_types/struct.TransactionWithStatus.html

blckngm (approved):

jjyr:
> > fix get_transaction_status throw invalid type: null, expected struct TransactionView error for rejected tx, return Ok(None) instead.
> 
> It seems that this could be fixed by updating ckb-jsonrpc-types too: https://docs.rs/ckb-jsonrpc-types/latest/ckb_jsonrpc_types/struct.TransactionWithStatus.html

Correct, we should also upgrade ckb_sdk to new version.

PR #684, opened 2022-05-09
Feat package failed transactions

jjyr:
The behavior of only package success transactions caused some problems:

1. Metamask increases sender's nonce when a transaction fail wether Godwoken packages the transaction. So 
the nonce becomes inconsistent once user send a failed transaction.
2. DOS attacking may happen if we do not charging fee for failed transactions. There are other ways to fix DOS, but charging fee for failed transaction is the easiest one to implement. To implement this behavior also solve the compatibility issues for some EVM toolchains.

So this PR introduce a backward compatible way to fix this. When a transaction is failed(return a non-zero exit code), the written state of the transaction is reverted, the sender's nonce is increased by 1, and the transaction fee is paid to block-producer. For `PolyjuiceTx` the `gas_used * gas_price` amount of CKB is charged. For other transaction, the `args.fee` is charged.

> Notice, we should update the readonly node first, then to update the sequencer, otherwise the readonly node may stop to work.


crates/jsonrpc-types/src/godwoken.rs
```diff
@@ -139,6 +139,7 @@ pub struct TxReceipt {
     pub post_state: AccountMerkleState,
     pub read_data_hashes: Vec<H256>,
     pub logs: Vec<LogItem>,
+    pub exit_code: Uint32,
```
blckngm, line 142: Should RPC doc be updated?

jjyr: Updated

blckngm (approved):

zeroqn (approved):

PR #687, opened 2022-05-10
fix(mempool): pending withdrawals not in mem block aren't restored

zeroqn:
- restore pending withdrawals from db

### DIFF
--- a/crates/tests/src/tests/restore_mem_pool_pending_withdrawal.rs
+++ b/crates/tests/src/tests/restore_mem_pool_pending_withdrawal.rs
@@ -92,11 +92,13 @@ async fn test_restore_mem_pool_pending_withdrawal() {
             })
             .collect()
     };
-    let pending_withdrawals = withdrawals.split_off(withdrawals.len() - 5);
+    let pending_withdrawals = withdrawals.split_off(withdrawals.len() - 10);
     let mem_block_withdrawals = withdrawals;
+    assert!(!pending_withdrawals.is_empty());
 
     // Insert error nonce withdrawal and expect them to be removed during pending restore
     let invalid_withdrawals_count = pending_withdrawals.len() - 5;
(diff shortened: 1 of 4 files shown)


jjyr (dismissed):

Flouse:
What is the plan of `TODO: Check replace-by-fee` in `fn push_withdrawal_request`?
@jjyr

crates/tests/src/tests/restore_mem_pool_pending_withdrawal.rs
```diff
@@ -0,0 +1,235 @@
...
+                    .owner_lock(owner_lock)
+                    .build()
+            })
+            .collect()
+    };
+    let pending_withdrawals = withdrawals.split_off(withdrawals.len() - 5);
+    let mem_block_withdrawals = withdrawals;
+
+    // Insert error nonce withdrawal and expect them to be removed during pending restore
+    let invalid_withdrawals_count = pending_withdrawals.len() - 5;
```
Flouse, line 100: `invalid_withdrawals_count` here would be zero?

Flouse (changes requested):

crates/tests/src/tests/restore_mem_pool_pending_withdrawal.rs
zeroqn: This count is used to create invalid withdrawals for test, should not be zero.

zeroqn: fixed

Flouse (approved):

blckngm (approved):

jjyr:
> What is the plan of `TODO: Check replace-by-fee` in `fn push_withdrawal_request`? @jjyr

We should implement a similar mechanism like TxFeeQueue for the withdrawal

PR #690, opened 2022-05-12
fix: rpc client report not enough finalized sudt custodians

zeroqn:
- fix: collect finalized ckb custodians more than is required
- fix: finalized ckb custodians aren't merged. `output_data_len_range` takes `[inclusive, exclusive]` and enforce empty data for ckb deposit.
- fix: withdrawal generate change ckb custodian with u64::MAX capacity. `output_capacity_range` takes `[inclusive, exclusive]`.

### DIFF
--- a/crates/rpc-client/src/rpc_client.rs
+++ b/crates/rpc-client/src/rpc_client.rs
@@ -1051,7 +1051,7 @@ impl RPCClient {
             script: None,
             block_range: None,
             output_data_len_range: None,
-            output_capacity_range: Some([min_capacity.into(), u64::MAX.into()]),
+            output_capacity_range: Some([min_capacity.into(), u64::MAX.into()]), // [inclusive, exclusive]
         });
         let search_key = SearchKey {
             script: ckb_types::packed::Script::new_unchecked(custodian_lock.as_bytes()).into(),
--- a/crates/rpc-client/src/rpc_client.rs
+++ b/crates/rpc-client/src/rpc_client.rs
@@ -1074,6 +1074,14 @@ impl RPCClient {
                     return Ok(QueryResult::NotEnough(collected));
                 }
 
+                // Skip ckb custodians if capacity is fullfill
+                if collected.capacity >= required_capacity
+                    && !withdrawals_amount.sudt.is_empty()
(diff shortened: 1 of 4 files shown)


crates/rpc-client/src/rpc_client.rs
```diff
@@ -996,15 +1002,36 @@ impl RPCClient {
...
+            min_capacity,
+            max_cells,
+        )
+        .await
+    }
+
+    // TODO: refactor into custodian mod
+    #[instrument(skip_all)]
+    async fn query_finalized_custodian_cells_from_indexer(
+        indexed_cells: &impl IndexedCells,
```
blckngm, line 1021: It's a bit strange having a testing injection point here. Could it be pushed to lower layers, e.g. in `request`?

blckngm: It is feasible to add some `cfg(test)` only code to `CKBIndexerClient` for mocking responses, e.g. `set_mock_response(response_callback)`. This way you won't need the `IndexedCells` trait and this `query_finalized_custodian_cells_from_indexer` method, and other RPC methods can be more easily tested too.

zeroqn: I think standalone mock struct for test only is more flexible. I plan to refactor all custodian query code into custodian mod. It doesn't need to care about how to construct `CKBIndexerClient`.

Then in `rpc_client`, we simply call ```custodian::query_finalized_cells``` instead. Or ```CustodianCells::new(impl xxx).query_finalized```.

blckngm: OK then.

blckngm (approved):

crates/rpc-client/src/rpc_client.rs
```diff
@@ -490,6 +491,11 @@ impl RPCClient {
             });
 
             for cell in cells {
+                // Ensure finalized ckb custodians are clearly mergeable
+                if cell.output.type_().is_none() && !cell.data.is_empty() {
```
jjyr, line 495: Why reject only CKB deposits?

jjyr (approved):

PR #696, opened 2022-05-18
Doc: deposit and withdrawal

jjyr:


docs/deposit_and_withdrawal.md
```diff
@@ -0,0 +1,102 @@
+# Deposit and withdrawal
+
+Deposit and withdrawal is a special layer1 <-> layer2 messaging mechanism with assets transfer.
```
blckngm: Make it clear that deposit is L1 -> L2, while withdrawal is L2 -> L1?

Users can deposit assets from layer1 (CKB) to layer2 (godwoken), or withdraw from layer2 back to layer1.

docs/deposit_and_withdrawal.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Deposit
+
+User deposit by sending a layer1 transaction on CKB while generating a cell with a special lock - deposit lock, which allows the block producer to unlock the cell. The block producer unlocks the deposited cell and then moves the assets under the custodian lock, at the same time, the block producer writes the deposit assets to layer2 state. All the process is enforced by layer1 script, even the block producer can't take the assets away.
```
blckngm:
```suggestion
A deposit request is a layer1 transaction which generates a cell with a special lock - deposit lock. The block producer will collect these cells and process the deposit requests in blocks. It will unlock the deposit cells, move the assets under the custodian lock, and update layer2 state, in block submission layer1 transactions. These transactions are checked by a layer1 script, so the block producer can't take the assets away.
```

docs/deposit_and_withdrawal.md
```diff
@@ -0,0 +1,102 @@
...
+    cancel_timeout: Uint64,
+    registry_id: Uint32,
+}
+```
+
+`DepositLockArgs` denotes the layer2 account's script and `registry_id`, currently, only the ETH registry is supported. Users can cancel the deposit after `cancel_timeout`, it is used in case the block producer rejects to package the deposited cell, it happened when the deposited cell contains invalid data.
+
+## Custodian cell
+
+The deposit cell is converted into a custodian cell, which means the assets are locked by layer2, custodian lock enforced the safety of the custodian assets, the assets can only be transferred out when the user withdrawal.
```
blckngm:
```suggestion
Deposit cells are converted to custodian cells when assets are deposited to layer2. Custodian cells are protected by the custodian lock, which enforces that the assets can only be transferred out when a user withdraw.
```

docs/deposit_and_withdrawal.md
```diff
@@ -0,0 +1,102 @@
...
+
+CKB requires `capacity` to cover the cost of the cell, the `capacity` of the deposited cell must also cover the custodian cell, so the minimal deposit CKB that Godwoken allows is as follows:
+
+* Deposit CKB: 246 CKB
+* Deposit CKB and Simple UDT: 327 CKB
+
+
+## Withdrawal
+
+Withdrawal is the reverse process of the deposit. A user signs a withdrawal request and sends it to the block producer, block producer converts the custodian cell into a withdrawal cell and updates the layer2 state.
```
blckngm:
```suggestion
Users sign withdrawal requests and send them to the block producer. The block producer will process these requests in blocks, update layer2 state and convert custodian cells to withdrawal cells in block submission layer1 transactions.
```

blckngm (commented):
Explain that deposit and withdrawal requests are processed in blocks, and cells are generated in block submission transactions.

docs/deposit_and_withdrawal.md
```diff
@@ -0,0 +1,102 @@
+# Deposit and withdrawal
+
+Deposit and withdrawal is a special layer1 <-> layer2 messaging mechanism with assets transfer. Users can deposit assets from layer1 (CKB) to layer2 (godwoken), or withdraw from layer2 back to layer1.
+
+## Deposit
+
```
keroro520, line 6: A deposit request is a layer1 cell but not a transaction

keroro520: deposit lock =＞ deposit-lock

And better add the link to deposit-lock implementation(gw-scripts)

blckngm (dismissed):

docs/deposit_and_withdrawal.md
```diff
@@ -0,0 +1,102 @@
...
+lock:
+  code_hash:    (withdrawal lock's code hash),
+  hash_type:    Type,
+  args: (rollup_type_hash(32 bytes) | WithdrawalLockArgs (n bytes) | len (4 bytes) | layer1 owner lock (n bytes))
+capacity:   (CKB amount),
+type_:  (none or SUDT script)
+data:   (none or SUDT amount)
+```
+
+Withdrawal lock guarantees the user can unlock this cell after `finality blocks`.
```
keroro520, line 85: Maybe we can remove things related to finality blocks out of this doc?

Otherwise please explain more detail about the relation between finality blocks and withdrawal, and what if there are not enough finalized custodian cells for a withdrawal.

keroro520 (changes requested):

docs/deposit_and_withdrawal.md
blckngm: > A deposit request is a layer1 cell but not a transaction

How about “A deposit request is created by sending a layer1 transaction which ...”

docs/deposit_and_withdrawal.md
jjyr: > what if there are not enough finalized custodian cells for a withdrawal.

IMO, we can add the content in anytime, but these missing parts does not prevent users to understand the current document.  I will add a link to explain `finality blocks` after we had that document.

keroro520: Sure

docs/deposit_and_withdrawal.md
jjyr: Update the description.

I think we should use 'deposit lock' rather than 'deposit-lock', lock is a concept of CKB, the deposit lock implies the script is the `lock` field of a CKB cell. But the 'deposit-lock' implies it is a unique concept not related to the CKB's lock, which isn't true.

keroro520: Both are okay to me

keroro520 (approved):

zeroqn (approved):

PR #702, opened 2022-05-25
fix/refactor some CKB/godwoken RPC related code

blckngm:
Fixed `get_transaction` CKB RPC response parsing, and `get_registry_address` godwoken RPC response parsing.

Refactored “get CKB transaction”/“wait CKB transaction” related code.

Switched to `tracing-subscriber` in tools too because `env_logger` gets quite noisy with all the traces.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2324,7 +2324,6 @@ dependencies = [
  "ckb-types",
  "clap",
  "csv",
- "env_logger",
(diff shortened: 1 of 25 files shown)


keroro520 (approved):

jjyr (approved):

blckngm:
Dispatch from PR does not seem to actually run the integration test?

/itest

PR #703, opened 2022-05-25
perf: remove reduandent checking deposits expired

keroro520:


crates/mem-pool/src/pool.rs
```diff
@@ -783,34 +783,39 @@ impl MemPool {
...
-                    force_expired = true;
-                    break;
-                }
             }
         }
 
+        if force_expired {
+            log::debug!("[mem-pool] forced clear pending deposits");
+            self.pending_deposits.clear();
+        }
```
keroro520, line 817: Note for reviewers, when `force_expired` is `true`, we should clear `self.pending_deposits` before making the `safe_expired` prediction. Otherwise `safe_expired` will always be `false` when `force_expired` is equal to `true`.

blckngm (approved):

zeroqn (approved):

jjyr (approved):

PR #704, opened 2022-05-25
chore: increase timeout of gw-tools waiting layer1 tx committed

keroro520:
Why: CKB runs on poor CI machine may take too long to execute txs. Here
is an example: https://github.com/RetricSu/godwoken-kicker/runs/6586127378?check_suite_focus=true#step:15:50


Flouse (dismissed):

zeroqn (dismissed):

blckngm:
kicker deposit 没有用 `MANUAL_BUILD_GODWOKEN`...

jjyr (approved):

PR #706, opened 2022-05-30
feat: enrich WithdrawalWithStatus

keroro520:


crates/jsonrpc-types/src/godwoken.rs
```diff
@@ -541,6 +541,12 @@ impl Default for WithdrawalStatus {
 pub struct WithdrawalWithStatus {
     pub withdrawal: Option<WithdrawalRequestExtra>,
     pub status: WithdrawalStatus,
+    pub l1_block_number: Option<u64>,
```
jjyr: I suggest we can use a nested structure to represent comission information.
```
struct CommittedInfo {
    pub l1_block_number: Option<u64>,
    pub l1_block_hash: Option<H256>,
    pub l1_transaction_hash: Option<H256>,
    pub l2_block_number: Option<u64>,
    pub l2_block_hash: Option<H256>,
    pub l2_withdrawal_index: Option<u32>
}
```

jjyr (changes requested):

jjyr:
1. need to update doc/RPC.md
2. please rebase develop branch to pass the CI

crates/jsonrpc-types/src/godwoken.rs
keroro520: Without reasonable benefit, I prefer the flat structure as it is clearer than nested.

jjyr: After revisit I think we should separate these fields into the following two structure:

```
struct L2BlockInfo {
    pub l2_block_number: Option<u64>,
    pub l2_block_hash: Option<H256>,
    pub l2_withdrawal_index: Option<u32>
}
struct CommittedInfo {
    pub l1_block_number: Option<u64>,
    pub l1_block_hash: Option<H256>,
    pub l1_transaction_hash: Option<H256>,
}
```

The fields should grouped by the inter-relationship, so user can easily understand when a field is exist and when it is null. Even in the flat structure which you proposed the fields are grouped by their prefix name.

blckngm: Why not factor out the `Option`s?

```rust
struct WithdrawalWithStatus {
    ...
    pub l2_block_info: Option<L2BlockInfo>,
    pub l1_committed_info: Option<L1CommittedInfo>,
}
```

blckngm: We can also reuse `L2BlockCommittedInfo` for `l1_committed_info`.

keroro520: well, if we intend to use nested structure to illustrate the inter-relationship between these fields, the result structure will look like this:

```rust
pub struct WithdrawalWithStatus {
    pub withdrawal: Option<WithdrawalRequestExtra>,
    pub status: WithdrawalStatus,

    pub l1_committed_info: Option<L2BlockCommittedInfo>,
    pub l2_committed_info: Option<L2WithdrawalCommittedInfo>,
}

pub struct L2WithdrawalCommittedInfo {
    pub block_number: u64,
    pub block_hash: H256,
    pub withdrawal_index: u32
}
```

jjyr (dismissed):

blckngm:
RPC doc should be updated.

keroro520:
Sorry so long to reply.

Updated:
- d76247e changes to the nested structure
- 19f4ae1 updates the RPC document

@sopium @jjyr PTAL

blckngm (approved):

zeroqn (approved):

PR #709, opened 2022-05-30
fix: don't set TCP keepalive on listening sockets

blckngm:
On Mac OS X, it causes the listening socket to be “closed” after the
keepalive timeout and future connections to fail with “connection
refused” errors. We suspect the same may happen on linux in some
circumstances because we see similar errors in CI.

For hyper, we switch to use `AddrIncoming::set_keepalive`. It sets TCP
keepalive options on sockets returned by `accept` instead of on the
listening socket.

Tentacle unfortunately does not seem to allow setting options on
accepted sockets, so TCP keepalive setting is removed for now.

### DIFF
--- a/crates/rpc-server/src/server.rs
+++ b/crates/rpc-server/src/server.rs
@@ -22,21 +22,16 @@ pub async fn start_jsonrpc_server(
     let rpc_server = registry.build_rpc_server()?;
 
     let listener = TcpListener::bind(listen_addr).await?;
-    let listener_ref = socket2::SockRef::from(&listener);
-    // Set TCP keepalive options with socket2 because tokio/hyper does not support setting interval/retries (yet).
-    let keepalive = socket2::TcpKeepalive::new()
-        .with_time(Duration::from_secs(10))
-        .with_interval(Duration::from_secs(5))
-        .with_retries(3);
-    // TCP keepalive options set on listening sockets are inhereted by accepted sockets (at least on linux and FreeBSD).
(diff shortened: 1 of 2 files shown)


zeroqn (approved):

crates/rpc-server/src/server.rs
```diff
@@ -22,21 +22,16 @@ pub async fn start_jsonrpc_server(
...
 
     // Format the full address.
     let url = format!("http://{}", listener.local_addr()?);
     log::info!("JSONRPC server listening on {}", url);
 
+    let mut incoming = AddrIncoming::from_listener(listener)?;
+    incoming.set_keepalive(Some(Duration::from_secs(10)));
+
     // Start a hyper server.
-    let server = Server::builder(AddrIncoming::from_listener(listener)?)
```
keroro520, line 39: https://github.com/hyperium/hyper/blob/f8e2a831943e9a687a1b3277c806caca6e0ac8a4/src/server/tcp.rs#L48-L58

```rust
    /// Creates a new `AddrIncoming` from an existing `tokio::net::TcpListener`.
    pub fn from_listener(listener: TcpListener) -> crate::Result<Self> {
        let addr = listener.local_addr().map_err(crate::Error::new_listen)?;
        Ok(AddrIncoming {
            listener,
            addr,
            sleep_on_errors: true,
            tcp_keepalive_timeout: None,
            tcp_nodelay: false,
            timeout: None,
        })
    }
```

curious why hyper doesn't inherit the keepalive options from socket

keroro520 (approved):

PR #713, opened 2022-05-31
feat: support switch backend binaries by height

jjyr:
This PR support upgrade backend binaries by height.

Now, instead of define `backends` in the `config.toml` we define `backend_switch`:

``` rust
pub struct BackendSwitchConfig {
    pub switch_height: u64,
    pub backends: Vec<BackendConfig>,
}
```

`switch_height` means in which height we use the new version of backend. For example:

To upgrade godwoken config file, we should move all the `backends` fields into a backend switch clause such as: `BackendSwitchConfig {switch_height: 0, backends}`, when we need to upgrade a backend, we define a new backend switch item follows the original one such as: `BackendSwitchConfig {switch_height: 1000, backends: <new SUDT backend>}`. Then when the tip block arrives the 1000 height, the new SUDT backend will be loaded instead the old one, the other backends are not change.

This change affects the following API:

https://github.com/nervosnetwork/godwoken/blob/develop/docs/RPC.md#method-gw_get_node_info

Consider mostly use-case of this API is to use the `validator_script_type_hash ` field of a backend, I think we can only returns backends at 0 height for now.

### DIFF
--- a/crates/generator/src/backend_manage.rs
+++ b/crates/generator/src/backend_manage.rs
@@ -95,12 +95,12 @@ impl BackendManage {
     fn compile_backend(&mut self, backend: &Backend) {
         self.aot_codes.0.insert(
             backend.validator_script_type_hash,
-            self.aot_compile(backend.generator, 0)
+            self.aot_compile(&backend.generator, 0)
                 .expect("Ahead-of-time compile"),
         );
         self.aot_codes.1.insert(
             backend.validator_script_type_hash,
-            self.aot_compile(backend.generator, 1)
+            self.aot_compile(&backend.generator, 1)
                 .expect("Ahead-of-time compile"),
         );
     }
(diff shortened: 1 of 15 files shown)


zeroqn (approved):

crates/generator/src/backend_manage.rs
```diff
@@ -17,72 +17,107 @@ pub struct Backend {
...
-        self.backends
-            .insert(backend.validator_script_type_hash, backend);
+    pub fn get_backends_at_height(
+        &self,
+        block_number: u64,
+    ) -> Option<&(u64, HashMap<H256, Backend>)> {
+        self.backend_switches
+            .iter()
+            .rev()
+            .find(|(height, _)| block_number >= *height)
```
blckngm, line 115: [Binary search](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.binary_search_by_key) could be used to improve performance if the number of switches is high.

blckngm (approved):

crates/generator/src/backend_manage.rs
jjyr: Right, but since the chain is keep growing, the later version of backend has more possibility to be selected.

PR #721, opened 2022-06-08
fix: refresh readonly mem-pool when receives new mem-block

jjyr:
The Readonly node should not do it's own computation when the mem-pool reset, instead, the node should always wait for the full node to publish a new mem-block state, and then to sync with that state.

In this PR:

1. Fix the inconsistent state of readonly node by stop doing reset in the `MemPool#reset` function.
2. ReadOnly node cleans the mem-pool state when receives a new mem-block from fullnode.
3. Fix filter_withdrawals, prevent removing the follow-up withdrawals from a same sender.

**Note: The web3-indexer should subscribe the readonly node API to keep consistent state.**

### DIFF
--- a/crates/mem-pool/src/pool.rs
+++ b/crates/mem-pool/src/pool.rs
@@ -664,44 +664,59 @@ impl MemPool {
             withdrawals.extend(mem_block_withdrawals);
         } else {
             // packages more withdrawals
-            fn filter_withdrawals(
-                state: &MemStateTree<'_>,
-                withdrawal: &WithdrawalRequestExtra,
-            ) -> bool {
-                let id = state
-                    .get_account_id_by_script_hash(&withdrawal.raw().account_script_hash().unpack())
-                    .expect("get id")
-                    .expect("id exist");
-                let nonce = state.get_nonce(id).expect("get nonce");
-                let expected_nonce: u32 = withdrawal.raw().nonce().unpack();
-                // ignore withdrawal mismatch the nonce
-                nonce == expected_nonce
-            }
-            withdrawals.retain(|w| filter_withdrawals(&mem_state, w));
-
-            // package withdrawals
-            if withdrawals.len() < self.mem_block_config.max_withdrawals {
-                for entry in self.pending().values() {
-                    if let Some(withdrawal) = entry.withdrawals.first() {
-                        if filter_withdrawals(&mem_state, withdrawal) {
-                            withdrawals.push(withdrawal.clone());
-                        }
-                        if withdrawals.len() >= self.mem_block_config.max_withdrawals {
-                            break;
-                        }
-                    }
-                }
-            }
+            self.try_package_more_withdrawals(&mut mem_state, &mut withdrawals);
         }
 
-        self.prepare_next_mem_block(&db, &mut mem_state, withdrawals, txs)
-            .await?;
+        self.prepare_next_mem_block(
+            &db,
+            &mut mem_state,
+            withdrawals,
+            self.pending_deposits.clone(),
+            txs,
+        )
+        .await?;
 
         // store mem state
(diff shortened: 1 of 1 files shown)


crates/mem-pool/src/pool.rs
```diff
@@ -1181,19 +1208,28 @@ impl MemPool {
...
-        self.finalize_withdrawals(&mut state, withdrawals).await?;
-        self.finalize_deposits(&mut state, deposits).await?;
+        let mem_store = MemStore::new(snapshot);
+        mem_store.update_mem_pool_block_info(self.mem_block.block_info())?;
+        let mut mem_state = mem_store.state()?;
+
+        // remove from pending
+        let db = self.store.begin_transaction();
+        self.remove_unexecutables(&mut mem_state, &db).await?;
+        self.finalize_withdrawals(&mut mem_state, withdrawals)
```
blckngm: `finalize_withdrawals` may fail if read only node cannot collect enough finalized custodians, because e.g. it calls `finalize_withdrawals` later than the full node and some finalized custodian cells are now dead.

Do we want to skip collecting custodian cells in read only nodes?

jjyr: You are right, we should stop collecting custodian cells for readonly node.

crates/mem-pool/src/pool.rs
```diff
@@ -445,11 +445,47 @@ impl MemPool {
...
+    ) -> Result<()> {
+        let new_tip = match new_tip {
+            Some(block_hash) => block_hash,
+            None => {
+                log::debug!("reset new tip to last valid tip block");
+                self.store.get_last_valid_tip_block_hash()?
+            }
+        };
+        let new_tip_block = self.store.get_block(&new_tip)?.expect("new tip block");
+        self.current_tip = (new_tip, new_tip_block.raw().number().unpack());
```
blckngm, line 481: Need to broadcast that the tip is updated, or p2p sync will stuck on “waiting for tip update”.

```rust
        // Publish new tip.
        let _ = self.new_tip_publisher.send(self.current_tip);
```

crates/mem-pool/src/pool.rs
```diff
@@ -445,11 +445,53 @@ impl MemPool {
...
+
+        // Publish new tip.
+        let _ = self.new_tip_publisher.send(self.current_tip);
+        if let Some(ref publish) = self.mem_pool_publish_service {
+            publish.new_tip(self.current_tip).await;
```
blckngm: This is only necessary on full node for now. Because we do not support cascading sync (read only node syncing from another read only node) yet.

blckngm:
/itest
kicker: refs/pull/276

blckngm:
/itest
kicker: refs/pull/276/merge

blckngm:
/itest
kicker: 71fd3dd03345a5f0a1479736324144d7644ac084

Used outdated commit. Ignore this.

blckngm (dismissed):

sopium:
/itest
kicker: 16c5b6e50a2a198368273a7a2418337e27399eaf

sopium:
/itest
kicker: 16c5b6e50a2a198368273a7a2418337e27399eaf

zeroqn (dismissed):

blckngm:
/itest
godwoken: a6ec0d9ddf4dd658c797b9f8061899ac6b3cf9f0
kicker: bdc8b31163b03b80ededfb5a0b52ed1bde28eb40

Flouse (dismissed):

jjyr:
/itest
kicker: bdc8b31163b03b80ededfb5a0b52ed1bde28eb40

Flouse (approved):

blckngm (approved):

PR #722, opened 2022-06-11
Fix backends aot_code loading

jjyr:
- https://github.com/nervosnetwork/godwoken/pull/713 introduce a bug: 

If we setup a backend upgrade, the later version's aot binary will overwrite the previous one.

This PR fixed the bug, and introduce more logs of backend_manage module.


Flouse (commented):
approved

PR #725, opened 2022-06-13
fix: reject transactions has less gas than the intrinsic gas

jjyr:
Related PR in Polyjuice
- https://github.com/nervosnetwork/godwoken-polyjuice/pull/160

### DIFF
--- a/crates/utils/src/polyjuice_parser.rs
+++ b/crates/utils/src/polyjuice_parser.rs
@@ -1,5 +1,8 @@
 use gw_types::{bytes::Bytes, packed::RawL2Transaction, prelude::*};
 
+/// The data structure of the Polyjuice transaction arguments
+///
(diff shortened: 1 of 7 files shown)


magicalne (dismissed):

crates/utils/src/polyjuice_parser.rs
```diff
@@ -0,0 +1,55 @@
+use gw_types::{bytes::Bytes, packed::RawL2Transaction, prelude::*};
+
```
Flouse, line 2:
```suggestion

/// The data structure of the Polyjuice transaction arguments
///
/// see: https://github.com/nervosnetwork/godwoken-polyjuice/blob/main/README.md#polyjuice-arguments
```

Flouse (dismissed):

Flouse (approved):

PR #726, opened 2022-06-14
feat(rpc-server): get in queue requests

blckngm:
Now on a full node the `gw_get_transaction`/`gw_get_withdrawal` RPC
calls will return status: pending instead of null for in queue
transactions/withdrawals.

Also added RPC method `gw_is_request_in_queue`, which returns whether a
transaction or withdrawal is in queue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2030,15 +2030,11 @@ name = "gw-rpc-server"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "async-channel",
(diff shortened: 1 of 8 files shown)


jjyr:
Only full node has fee queue, readonly node can skip query from the map, can we pass `InQueueRequestMap ` as a `Option<T>` args to functions.

jjyr (approved):

zeroqn (approved):

PR #729, opened 2022-06-15
fix: handle VM InvalidCycles by generate a polyjuice system log

jjyr:
* Handle `ckb_vm::error::Error::InvalidCycles` error raise form CKB VM
* Set `run_result.exit_code` to `1` in that case.
* Generate a polyjuice system log `account: raw_tx.to_id, gas_used: gas, cumulative_gas_used: gas, created_address: [0u8; 20], status: 0,`.

### DIFF
--- a/crates/generator/src/generator.rs
+++ b/crates/generator/src/generator.rs
@@ -29,7 +29,6 @@ use gw_common::{
     },
     H256,
 };
-use gw_config::BackendType;
 use gw_store::{state::state_db::StateContext, transaction::StoreTransaction};
 use gw_traits::{ChainView, CodeStore};
 use gw_types::{
@@ -133,8 +132,6 @@ impl Generator {
         let mut run_result = RunResult::default();
         let used_cycles;
         let exit_code;
-        let mut appended_log = None;
-
         {
             let t = Instant::now();
             let global_vm_version = GLOBAL_VM_VERSION.load(SeqCst);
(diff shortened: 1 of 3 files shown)


classicalliu (dismissed):

zeroqn (dismissed):

crates/generator/src/generator.rs
```diff
@@ -170,8 +174,34 @@ impl Generator {
             let mut machine = TraceMachine::new(default_machine);
 
             machine.load_program(&backend.generator, &[])?;
-            exit_code = machine.run()?;
-            used_cycles = machine.machine.cycles();
+            match machine.run() {
+                Ok(_exit_code) => {
```
Flouse, line 175: If `run_polyjuice()` return before [`emit_evm_result_log` function](https://github.com/nervosnetwork/godwoken-polyjuice/blob/1.2-rc/c/polyjuice.h#L1384),
Polyjuice may return an `exit_code` without `POLYJUICE_SYSTEM_LOG` too.

Flouse (changes requested):

Flouse (dismissed):

crates/generator/src/generator.rs
Flouse: https://github.com/nervosnetwork/godwoken/pull/729/commits/4d7112652da9c5a5a065f37b486b67a0caeae765

Flouse (approved):

magicalne (approved):

PR #732, opened 2022-06-15
feat(block-producer): trace chain task run func

zeroqn:
- trace all ```handle_event``` funcs
- trace ```interval sleep``` and ```backoff sleep```

### DIFF
--- a/crates/block-producer/src/runner.rs
+++ b/crates/block-producer/src/runner.rs
@@ -872,6 +872,8 @@ pub async fn run(config: Config, skip_config_check: bool) -> Result<()> {
             let shutdown_send = shutdown_send.clone();
             move || {
                 rt_handle.block_on(async move {
+                    use tracing::Instrument;
+
                     let _tx = chain_task_ended_tx;
                     let ctx = ChainTaskContext {
                         chain_updater,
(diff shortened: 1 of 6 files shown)


jjyr (dismissed):

crates/block-producer/src/runner.rs
```diff
@@ -875,13 +888,45 @@ pub async fn run(config: Config, skip_config_check: bool) -> Result<()> {
...
+                        // Exit if shutdown event is received.
+                        if chain_task.shutdown_event.try_recv().is_ok() {
+                            log::info!("ChainTask existed successfully");
+                            return;
+                        }
+
+                        let chain_task_run_span = info_span!("chain_task_run");
+                        let _run_guard = chain_task_run_span.enter();
+
+                        match chain_task.run(&run_status).await {
```
blckngm: Attach `chain_task_run_span` to `chain_task.run(..)`?

E.g. `chain_task.run(&run_status).instrument(chain_task_run_span).await`.

See: https://docs.rs/tracing/latest/tracing/struct.Span.html#in-asynchronous-code

Flouse (dismissed):

crates/block-producer/src/runner.rs
zeroqn: Fixed

blckngm: On another look, `run` already has an `instrument` attribute, so this shouldn't be necessary?

blckngm (approved):

Flouse (approved):

PR #752, opened 2022-07-05
fix: Update parse log

classicalliu:


crates/utils/src/script_log.rs
```diff
@@ -36,18 +39,45 @@ pub enum GwLog {
...
+        if to_address_byte_size == 0 {
+            8
+        } else {
+            28
+        }
+    };
 
-    let mut to_address = [0u8; 20];
-    to_address.copy_from_slice(&data[21..41]);
+    let to_address = match RegistryAddress::from_slice(&data[start..end]) {
```
blckngm: Use `get(start..end)` to avoid panic?

crates/utils/src/script_log.rs
```diff
@@ -36,18 +39,45 @@ pub enum GwLog {
     },
 }
 
-fn parse_sudt_log_data(data: &[u8]) -> ([u8; 20], [u8; 20], u128) {
-    assert_eq!(data[0], 20);
-    let mut from_address = [0u8; 20];
-    from_address.copy_from_slice(&data[1..21]);
+fn parse_sudt_log_data(data: &[u8]) -> Result<(RegistryAddress, RegistryAddress, U256)> {
+    let mut start = 0;
```
jjyr, line 45: check length of data to avoid panic.

jjyr (changes requested):

crates/utils/src/script_log.rs
classicalliu: Updated

crates/utils/src/script_log.rs
classicalliu: Updated

jjyr (approved):

blckngm (approved):

PR #753, opened 2022-07-05
fix: always refresh pending deposit cells

jjyr:
The `refresh_deposit_cells` function is introduced in the https://github.com/nervosnetwork/godwoken/pull/433 , the original intension is to keep consistent of `account id` during the mem-pool refreshing, otherwise, the following txs will invalid  due to the incorrect signature.

In the v1 version, we migrate signing method to the Ethereum & EIP-712, the `account id` is not included in the signature anymore, even the `account id` is changed, the following txs still passes the validate.

### DIFF
--- a/devtools/fetch-binaries.sh
+++ b/devtools/fetch-binaries.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 dir_name='.tmp/binaries'
-image='ghcr.io/nervosnetwork/godwoken-prebuilds:1.2.2-rc1'
+image='ghcr.io/nervosnetwork/godwoken-prebuilds:1.3.0-rc3'
 
 docker pull $image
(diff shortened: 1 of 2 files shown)


zeroqn (dismissed):

keroro520 (dismissed):

devtools/fetch-binaries.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 dir_name='.tmp/binaries'
-image='ghcr.io/zeroqn/godwoken-prebuilds:v1.1-feat-change-ckb-decimal-to-18'
+image='ghcr.io/nervosnetwork/godwoken-prebuilds:1.2.2-rc1'
```
Flouse:
```suggestion
image='ghcr.io/nervosnetwork/godwoken-prebuilds:1.3.0-rc3'
```

Flouse (dismissed):

PR #754, opened 2022-07-06
feat: export and import block

zeroqn:
- add: `export-block` subcommand
  - example: ```godwoken export-block -c config.toml --output-path ./blocks_testnet_v1```
- add: `import-block` subcommand
  - example: ```godwoken import-block -c config.toml --source-path ./block_testnet_v1``` 
- test: `export-block` and `import-block`
- doc: `export-bock` and `import-block` [link](https://github.com/zeroqn/godwoken/blob/acf0cb833ccc139a465b07d7b0cd14ed34918824/docs/export_import_block.md)


crates/utils/src/export_block.rs
```diff
@@ -0,0 +1,257 @@
...
+    };
+
+    Ok(exported_block)
+}
+
+pub fn read_block(reader: &mut impl Read) -> Result<Option<(ExportedBlock, usize)>> {
+    let mut full_size_buf = [0u8; 4];
+    if let Err(err) = reader.read_exact(&mut full_size_buf) {
+        match err.kind() {
+            ErrorKind::UnexpectedEof if full_size_buf == [0u8; 4] => return Ok(None),
```
blckngm: This will also return `Ok(None)` if n zeros (n < 4) are read.

crates/utils/src/export_block.rs
```diff
@@ -0,0 +1,257 @@
...
+            _ => bail!(err),
+        }
+    }
+
+    let full_size = u32::from_le_bytes(full_size_buf) as usize;
+    if full_size <= 4 {
+        bail!("block corrupted, full size {}", full_size);
+    }
+
+    let mut buf = vec![0; full_size];
```
blckngm, line 115: Consider reusing the buffer? (Move it to `ExportedBlockReader`.)

crates/block-producer/src/subcommand/export_block.rs
```diff
@@ -0,0 +1,158 @@
+use std::fs;
```
jjyr: We should create a new `crates/cli` crate, and move cli related files such as `block-producer/src/main.rs` and `export_block.rs` to the new crate.

blckngm: May worth clarifying: what's the difference between `cli` and `tools`? What about `replay-chain`?

crates/utils/src/export_block.rs
zeroqn: This buffer will be moved in two places:
  - https://github.com/nervosnetwork/godwoken/pull/754/files#diff-c5d1ffc9e3eca2c0823859c004accb1afdfc0f535b10b73f33b65d3394857156R120
  - https://github.com/nervosnetwork/godwoken/pull/754/files#diff-07eff2ee69a5eac43558211b5b3efdfded30a207932dd888c54a52c2fc2f16b3R180

We can optimize later if this is a bottleneck.

crates/block-producer/src/subcommand/export_block.rs
zeroqn: `crates/godwoken-bin` crate ?

zeroqn: Extract code to `godwoken-bin` crate

crates/store/src/transaction/store_transaction.rs
```diff
@@ -103,10 +103,16 @@ impl StoreTransaction {
...
         &self,
         reverted_block_smt_root: &H256,
-        block_hashes: Vec<H256>,
+        prev_reverted_block_smt_root: H256,
+        mut block_hashes: Vec<H256>,
     ) -> Result<(), Error> {
         assert!(!block_hashes.is_empty(), "set empty reverted block hashes");
 
+        // Prefix block hashes with prev smt root, order of origin block hashes isn't a matter.
+        block_hashes.push(prev_reverted_block_smt_root);
```
jjyr, line 112: If we change the data, we should do it explicit, I suggest to change the return of `get_reverted_block_hashes` to a structure instead of `HashSet<H256>`.

zeroqn: Agree.  Add `SMTRevertedBlockHashes` struct.

```rust
pub struct SMTRevertedBlockHashes {
    pub prev_smt_root: H256,
    pub block_hashes: Vec<H256>,
}
```

blckngm (dismissed):

magicalne (dismissed):

jjyr:
Please resolve conflicts

jjyr (approved):

magicalne (approved):

PR #755, opened 2022-07-06
fix(block-producer): exit during bad block challenge

zeroqn:
- fix: no dump on dry run challenge tx failed
- fix: invalid code hash for challenger tx (omni lock)
- fix: chain notify mem pool with bad block tip hash cause mem pool set snapshot assertion failure
- fix: pool duplicated challenger tx due to replace poa lock with omni lock
- fix: mem pool isn't reset after bad block is reverted
- fix: one finalized custodian with no withdrawal

### DIFF
--- a/crates/block-producer/src/withdrawal.rs
+++ b/crates/block-producer/src/withdrawal.rs
@@ -37,7 +37,7 @@ pub fn generate(
     contracts_dep: &ContractsCellDep,
     withdrawal_extras: &HashMap<H256, WithdrawalRequestExtra>,
 ) -> Result<Option<GeneratedWithdrawals>> {
-    if block.withdrawals().is_empty() && finalized_custodians.cells_info.is_empty() {
+    if block.withdrawals().is_empty() && finalized_custodians.cells_info.len() <= 1 {
         return Ok(None);
     }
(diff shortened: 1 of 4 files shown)


crates/block-producer/src/withdrawal.rs
```diff
@@ -37,7 +37,7 @@ pub fn generate(
     contracts_dep: &ContractsCellDep,
     withdrawal_extras: &HashMap<H256, WithdrawalRequestExtra>,
 ) -> Result<Option<GeneratedWithdrawals>> {
-    if block.withdrawals().is_empty() && finalized_custodians.cells_info.is_empty() {
+    if block.withdrawals().is_empty() && finalized_custodians.cells_info.len() <= 1 {
```
jjyr, line 40: We can still generate withdrawal if there is only one finalize cell, why return none?

jjyr (approved):

blckngm (approved):

crates/block-producer/src/withdrawal.rs
zeroqn: Yes, but it's useless. To merge finalized custodian cells, we need at least two cells. We can safely return none here.

PR #758, opened 2022-07-13
Redirect syscall debug logs

magicalne:
Now we can send contract logs to `sentry` if we hit an error.
There are 3 modes In ContractLogConfig:
- Default - Just like the old times. And will not send to sentry.
- Redirect - Redirect logs to a spawned task and print them all.
- RedirectError - Redirect logs to a spawned task and only print logs if the exit code is not 0.

### DIFF
--- a/crates/config/src/config.rs
+++ b/crates/config/src/config.rs
@@ -45,6 +45,8 @@ pub struct Config {
     pub dynamic_config: DynamicConfig,
     #[serde(default)]
     pub p2p_network_config: Option<P2PNetworkConfig>,
+    #[serde(default)]
+    pub contract_log_config: ContractLogConfig,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
@@ -410,4 +412,18 @@ pub struct GithubConfigUrl {
 pub struct DynamicConfig {
     pub fee_config: FeeConfig,
     pub rpc_config: RPCConfig,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum ContractLogConfig {
+    Default,       //print contract log in place
+    Redirect,      //print all of the contract logs, send error logs to sentry
+    RedirectError, //only print logs when the tx hit an error, send error logs to sentry
+}
+
+impl Default for ContractLogConfig {
+    fn default() -> Self {
+        ContractLogConfig::Default
+    }
 }
--- a/crates/config/src/config.rs
+++ b/crates/config/src/config.rs
@@ -20,6 +20,7 @@ pub enum Trace {
 #[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
 pub struct Config {
     pub node_mode: NodeMode,
+    #[serde(default)]
     pub contract_log_config: ContractLogConfig,
     pub backend_switches: Vec<BackendSwitchConfig>,
     pub genesis: GenesisConfig,
--- a/crates/config/src/config.rs
+++ b/crates/config/src/config.rs
@@ -20,6 +20,7 @@ pub enum Trace {
 #[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
 pub struct Config {
     pub node_mode: NodeMode,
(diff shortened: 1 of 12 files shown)


crates/generator/src/syscalls/redir_log.rs
```diff
@@ -0,0 +1,195 @@
...
+            if exit_code != 0 {
+                let mut entries: Vec<String> = Vec::with_capacity(3);
+                if let Some(tx) = &self.tx {
+                    let tx_hash = hex::encode(tx.as_reader().hash());
+                    entries.push(tx_hash);
+                }
+                entries.push(s.to_string());
+                entries.push(format!("exit code: {}", exit_code));
+                let msg = entries.join("\n");
+                sentry::capture_message(&msg, sentry::Level::Error);
```
jjyr: We should use async version to send send message.

crates/generator/src/syscalls/redir_log.rs
```diff
@@ -0,0 +1,195 @@
...
+        self.tx = Some(tx);
+    }
+
+    fn append_log(&mut self, log: &[u8]) {
+        self.sink.put(log);
+        self.sink.put_u8(b'\n');
+    }
+
+    fn flush(&mut self, exit_code: i8) {
+        if let Ok(s) = std::str::from_utf8(&self.sink) {
```
jjyr: Use `log::error` to logging the error result.

crates/generator/src/syscalls/redir_log.rs
```diff
@@ -0,0 +1,195 @@
...
+            //  exit code: 3
+            if exit_code != 0 {
+                let mut entries: Vec<String> = Vec::with_capacity(3);
+                if let Some(tx) = &self.tx {
+                    let tx_hash = hex::encode(tx.as_reader().hash());
+                    entries.push(tx_hash);
+                }
+                entries.push(s.to_string());
+                entries.push(format!("exit code: {}", exit_code));
+                let msg = entries.join("\n");
```
blckngm: It should be possible to avoid allocation and concatenation? Push transaction hash and exit code to sink too and just `capture_msg(&s)` here.

crates/generator/src/syscalls/mod.rs
```diff
@@ -593,8 +599,7 @@ impl<'a, S: State, C: ChainView> L2Syscalls<'a, S, C> {
             addr += 1;
         }
 
-        let s = String::from_utf8(buffer).map_err(|_| VMError::ParseError)?;
-        log::debug!("[contract debug]: {}", s);
+        self.redir_log_handler.append_log(buffer);
```
blckngm: This seem to change the semantics of `output_debug`. Non-valid UTF-8 is now allowed. Is this intentional?

How about still convert to `String` here, and change `sink` to be a `String` too?

magicalne: makes sense

crates/generator/src/generator.rs
```diff
@@ -185,6 +193,7 @@ impl Generator {
                     return Err(err.into());
                 }
             }
+            self.redir_log_handler.flush(exit_code);
```
blckngm, line 196: If the function returns early, `flush` won't be called.

Also, could `machine_run` ever run concurrently?

magicalne: `machine_run` never runs concurrently. 
It's OK if this fn returns early. I think that only happens when the process exits. Plus, log printing is not that critical.

blckngm: If `flush` is not called, the next run will have logs of this run. A workaround is to clear the sink in setup instead flush.

crates/config/src/config.rs
```diff
@@ -20,6 +20,7 @@ pub enum Trace {
 #[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
 pub struct Config {
     pub node_mode: NodeMode,
+    pub contract_log_config: ContractLogConfig,
```
zeroqn, line 24: Add `#[serde(default)]`?
Without it will panic for exists config file.

jjyr (approved):

blckngm (approved):

zeroqn (approved):

PR #760, opened 2022-07-18
fix(rpc): calculate tx signature hash using packed bytes

zeroqn:
packed bytes contains size header, should use raw bytes.


zeroqn:
/itest
web3: 4d3dddc5881fedf0dbb8cce8a250aa423f9f5e99
scripts: master

Flouse (approved):

blckngm (approved):

jjyr (approved):

PR #772, opened 2022-07-28
Add RPC get_pending_tx_hashes

blckngm:
# New RPC `gw_get_pending_tx_hashes`

Get pending transaction hashes, which means a developer could  get the mem_pool transactions of Godwoken.


jjyr:
LGTM, can we turn this into 'Ready for review'?

classicalliu (approved):

classicalliu:
Does readonly node support this RPC ?

blckngm:
> Does readonly node support this RPC ?

Read-only node needs p2p sync to actually support this, or the list returned will always be empty.

jjyr (approved):

Flouse (approved):

PR #774, opened 2022-07-28
fix(tests): submit_withdrawal_request random failure

zeroqn:
Withdrawal request is pushed to mem pool immediately
https://github.com/nervosnetwork/godwoken/runs/7533962286?check_suite_focus=true#step:9:741

In this case, poll withdrawal until it's pushed to mem pool.

### DIFF
--- a/crates/tests/src/tests/rpc_server/submit_withdrawal_request.rs
+++ b/crates/tests/src/tests/rpc_server/submit_withdrawal_request.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use gw_common::{
     builtins::{CKB_SUDT_ACCOUNT_ID, ETH_REGISTRY_ACCOUNT_ID},
     ckb_decimal::CKBCapacity,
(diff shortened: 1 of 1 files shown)


crates/tests/src/tests/rpc_server/submit_withdrawal_request.rs
```diff
@@ -96,13 +96,15 @@ async fn test_submit_withdrawal_request() {
         .is_request_in_queue(withdrawal_hash)
         .await
         .unwrap();
-    assert!(is_in_queue);
 
-    chain.produce_block(vec![], vec![]).await.unwrap();
+    if !is_in_queue {
+        chain.produce_block(vec![], vec![withdrawal]).await.unwrap();
+    } else {
+        chain.produce_block(vec![], vec![]).await.unwrap();
```
blckngm: What if the withdrawal request remains in the queue for a while? The mem pool state subsequently loaded won't include it, right?

zeroqn: Maybe. By default `produce_block` fn will refresh mem pool, it will be included if it's pushed to mem pool. (aka `produce_block` fail to acquire lock)
Add new `produce_block_and_refresh_mem_pool` to allow skip refresh mem pool.

blckngm: My point is that this test is not waiting for the withdrawal to be actually pushed to the mem pool.

I think it should poll until the request is not in queue.

The latest commit still fails if I add a sleep before locking mem pool in line 602 in `registry.rs`.

zeroqn: Indeed. Fixed.

jjyr (dismissed):

blckngm (dismissed):

Flouse (dismissed):

jjyr (approved):

Flouse (approved):

blckngm (approved):

magicalne (approved):

PR #777, opened 2022-08-04
Support non EIP-155 tx

### LINKED ISSUE #775: Remove mandatory eip-155 txs for multi-chain factories deployment
[eip-2470](https://eips.ethereum.org/EIPS/eip-2470), [eip-1820](https://eips.ethereum.org/EIPS/eip-1820), [eip-4337](https://eips.ethereum.org/EIPS/eip-4337), [Gnosis factories](https://github.com/5afe/safe-factories), other smart contract wallets and some defi protocol (https://github.com/celo-org/celo-blockchain/issues/1734), all seem to rely on support for non-eip-155 transactions, especially eip-2470 and eip-1820, which are constructed rather than signed transactions, so no one knows the private key behind the transaction.

The strategy adopted by Godwoken now is to enforce the use of eip-155 transactions, which makes the above contracts difficult or impossible to deploy on Godwoken, which may affect the multi-chain deployment of some protocols.

Avax and celo used this strategy but changed it later due to various incompatibilities. https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0050.md

jjyr:
Related issue: #775 

To support EIP-155, Godwoken accepts `chain_id = 0` l2 txs now, we assume the tx is non-EIP-155 tx when `chain_id=0`.

Since we reuse the `chain_id` field, it implies if we deploy another Godwoken instance with `chain_id=0`, then we can't send a valid EIP-155 transaction on that chain since Godwoken assumes `chain_id=0` txs are non-EIP-155.

This PR also removes the unused Secp256K1 and Secp256K1Tron implementation.


jjyr:
@classicalliu Can you modify godwoken-indexer and web3 server to support non-EIP-155 txs

zeroqn (dismissed):

classicalliu (approved):

zeroqn (approved):

RetricSu (approved):

magicalne (approved):

PR #788, opened 2022-08-24
Validate native token transfer raw tx and signature

magicalne:
https://github.com/nervosnetwork/godwoken-polyjuice/pull/173

- check to_id == polyjuice_creator_id
- update `verify_tx` for native token transfer

### DIFF
--- a/crates/generator/src/generator.rs
+++ b/crates/generator/src/generator.rs
@@ -29,7 +29,8 @@ use gw_common::{
     },
     H256,
 };
-use gw_config::{BackendType, ContractLogConfig, SyscallCyclesConfig};
+
+use gw_config::{ContractLogConfig, SyscallCyclesConfig};
 use gw_store::{state::state_db::StateContext, transaction::StoreTransaction};
 use gw_traits::{ChainView, CodeStore};
 use gw_types::{
@@ -38,7 +39,7 @@ use gw_types::{
     offchain::{RollupContext, RunResult, RunResultCycles},
     packed::{
         AccountMerkleState, BlockInfo, ChallengeTarget, DepositRequest, L2Block, L2Transaction,
-        RawL2Block, RawL2Transaction, Script, TxReceipt, WithdrawalReceipt, WithdrawalRequestExtra,
+        RawL2Block, RawL2Transaction, TxReceipt, WithdrawalReceipt, WithdrawalRequestExtra,
     },
     prelude::*,
 };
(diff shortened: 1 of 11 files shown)


crates/utils/src/polyjuice_parser.rs
```diff
@@ -69,4 +74,8 @@ impl PolyjuiceParser {
             None
         }
     }
+
+    pub fn is_native_transfer(&self) -> bool {
+        self.is_call() && self.to_address().is_some()
```
keroro520, line 79: What about using `!self.is_create()`?
 
```
        !self.is_create() && self.to_address().is_some()
```

keroro520 (dismissed):

crates/utils/src/polyjuice_parser.rs
magicalne: Always accepts call: https://github.com/nervosnetwork/godwoken-polyjuice/pull/173#discussion_r946744062

zeroqn (dismissed):

crates/generator/src/generator.rs
```diff
@@ -736,6 +736,33 @@ impl Generator {
         &self.backend_manage
     }
 
+    pub fn get_polyjuice_creator_id<S: State + CodeStore>(
```
jjyr, line 747: This function can be defined in `utils` module since it is not related to generator's logic, we can pass the backend manage as the parameter.

crates/generator/src/verification/transaction.rs
```diff
@@ -1,29 +1,29 @@
...
+    generator: Arc<Generator>,
 }
 
 impl<'a, S: State + CodeStore> TransactionVerifier<'a, S> {
-    pub fn new(state: &'a S, rollup_context: &'a RollupContext) -> Self {
-        Self {
-            state,
-            rollup_context,
-        }
+    pub fn new(state: &'a S, generator: Arc<Generator>) -> Self {
```
jjyr: We can read `polyjuice_creator_id` from parameter to avoid duplicated DB read.

crates/generator/src/verification/transaction.rs
```diff
@@ -93,6 +93,17 @@ impl<'a, S: State + CodeStore> TransactionVerifier<'a, S> {
                 )
                 .into());
             }
+            // Native token transfer
+            if let Some(parser) = tx.parser() {
```
jjyr: return an error if failed to parse

jjyr (changes requested):

jjyr (dismissed):

crates/generator/src/verification/transaction.rs
```diff
@@ -93,6 +99,14 @@ impl<'a, S: State + CodeStore> TransactionVerifier<'a, S> {
                 )
                 .into());
             }
+            // Native token transfer
+            if p.is_native_transfer() {
+                // Verify to_id is CREATOR_ID
+                let to_id = raw_tx.to_id().unpack();
+                if Some(to_id) != self.polyjuice_creator_id && self.polyjuice_creator_id.is_some() {
```
zeroqn: if ```polyjuice_creator_id``` is ```None```, this check will pass? Caller must ensure ```polyjuice_creator_id``` is ```Some```?
Maybe pass ```beckend_manage``` and call ```utils::get_polyjuice_creator_id``` if ```polyjuice_creator_id``` is ```None```?

crates/mem-pool/src/pool.rs
```diff
@@ -454,6 +461,20 @@ impl MemPool {
         Self::package_mem_block(&self.mem_block, output_param)
     }
 
+    fn get_polyjuice_creator_id(&mut self) -> Result<Option<u32>> {
+        let snap = self.mem_pool_state.load();
+        let state = snap.state()?;
```
jjyr: Move these to line to the if branch, we only need the `state` when `if self.polyjuice_creator_id.is_none()` returns true.

jjyr (changes requested):

jjyr (approved):

keroro520 (approved):

PR #790, opened 2022-08-24
fix: recover sender of non EIP-155 txs

jjyr:
https://github.com/nervosnetwork/godwoken/pull/777

### DIFF
--- a/crates/types/src/extension.rs
+++ b/crates/types/src/extension.rs
@@ -51,6 +51,10 @@ impl packed::RawL2Transaction {
     pub fn hash(&self) -> [u8; 32] {
         self.as_reader().hash()
     }
+
+    pub fn is_chain_id_protected(&self) -> bool {
+        self.chain_id().unpack() != 0
+    }
 }
 
 impl packed::L2Transaction {
--- a/crates/generator/src/account_lock_manage/secp256k1.rs
+++ b/crates/generator/src/account_lock_manage/secp256k1.rs
@@ -43,7 +43,8 @@ impl Secp256k1Eth {
         receiver_script: &Script,
     ) -> anyhow::Result<H256> {
         let tx_chain_id = raw_tx.chain_id().unpack();
-        if chain_id != tx_chain_id {
+        let is_protected = raw_tx.is_chain_id_protected();
+        if is_protected && chain_id != tx_chain_id {
             bail!("mismatch tx chain id");
         }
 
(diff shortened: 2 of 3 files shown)


classicalliu (approved):

crates/types/src/extension.rs
```diff
@@ -51,6 +51,10 @@ impl packed::RawL2Transaction {
     pub fn hash(&self) -> [u8; 32] {
         self.as_reader().hash()
     }
+
+    pub fn is_chain_id_protected(&self) -> bool {
```
keroro520, line 55:
```suggestion
    pub fn is_eip155_tx(&self) -> bool {
```

I rename this function because I think "protected chain id" is difficult to understand, and "eip155" points to more useful information.

keroro520 (changes requested):

keroro520:
Which part of code **recovers the sender of non EIP-155 txs**? @jjyr

zeroqn:
> Which part of code **recovers the sender of non EIP-155 txs**? @jjyr

Only hash six rlp encoded items of tx for signing message 

https://github.com/nervosnetwork/godwoken/blob/6c4352d37233539b90403071fdd642a6cc6c808b/crates/generator/src/account_lock_manage/secp256k1.rs#L271-L276

keroro520:
> > Which part of code **recovers the sender of non EIP-155 txs**? @jjyr
> 
> Only hash six rlp encoded items of tx for signing message
> 
> https://github.com/nervosnetwork/godwoken/blob/6c4352d37233539b90403071fdd642a6cc6c808b/crates/generator/src/account_lock_manage/secp256k1.rs#L271-L276

Thanks a lot!

crates/types/src/extension.rs
jjyr: The `chain_id` should be interpreted by different EOA locks, since we only have the ETH EOA, it is correct to call it EIP-155. But actually the tx isn't encoded in EIP-155, it encoded in Godwoken L2TX format, so I prefer the original name.

zeroqn (approved):

keroro520 (approved):

crates/types/src/extension.rs
keroro520: Now LGTM. The core concept is "protect tx by chain id". my intention is to reduce the barrier of understanding by using the same concept.

PR #801, opened 2022-09-14
fix(mem-pool): fee queue use meta contract nonce for txs from id 0

zeroqn:
Fee queue will increase nonce from same account id, but txs from id 0 always set nonce to 0. This cause multiple txs are incorrectly dropped during `fetch`.

In this pr, we separate txs from id 0 by their signature hash, and correctly increase their nonces.

### DIFF
--- a/crates/mem-pool/src/fee/types.rs
+++ b/crates/mem-pool/src/fee/types.rs
@@ -13,6 +13,7 @@ use std::{cmp::Ordering, convert::TryInto};
 #[derive(PartialEq, Eq, Clone, Copy, Debug)]
 pub enum FeeItemKind {
     Tx,
+    PendingCreateSenderTx,
     Withdrawal,
 }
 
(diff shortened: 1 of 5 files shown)


crates/mem-pool/src/fee/types.rs
```diff
@@ -67,14 +75,20 @@ impl PartialOrd for FeeItem {
     }
 }
 
+#[derive(PartialEq, Eq, Clone, Copy, Debug, Hash)]
+pub enum FeeItemSender {
+    AccountId(u32),
+    PendingCreate(H256), // hash
```
blckngm, line 81: The hash here is never read? Is it for distinguishing different senders in the `fetched_senders` map?

blckngm (approved):

crates/mem-pool/src/fee/types.rs
zeroqn: Yes, for `fetched_senders` map.

RetricSu (approved):

jjyr (approved):

PR #807, opened 2022-09-29
refactor(generator): unchecked_execution_transaction and move it behind a feature gate

zeroqn:
For easily godwoken-scripts custom contracts test.  For example:
  - https://github.com/nervosnetwork/godwoken-scripts/blob/master/tests/src/script_tests/l2_scripts/examples.rs#L205

`unchecked_execute_transaction` and `execute_transaction` are identical. It also pays fee for failed tx. But for custom contract, which uses BackendType::Unknown, it always result in `Transaction::NoCost`.

Refactor `unchecked_execute_transaction` to `unhandled_execute_transaction`. It won't call `handle_run_result` and return origin `RunResult`.

Make this api behind `unhandled-execution` feature gate.

### DIFF
--- a/crates/generator/src/generator.rs
+++ b/crates/generator/src/generator.rs
@@ -698,7 +698,7 @@ impl Generator {
         max_cycles: u64,
         cycles_pool: Option<&mut CyclesPool>,
     ) -> Result<RunResult, TransactionError> {
-        let run_result = self.unchecked_execute_transaction(
+        let args = self.prepare_machine_run_args(
             chain,
             state,
             block_info,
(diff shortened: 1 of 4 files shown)


crates/generator/src/generator.rs
```diff
@@ -698,20 +698,24 @@ impl Generator {
...
-        Ok(run_result)
+
+        let run_result = self.machine_run(args)?;
+        self.handle_run_result(state, block_info, raw_tx, run_result)
     }
 
-    /// execute a layer2 tx, doesn't check exit code
+    /// execute a layer2 tx without handle run result
+    /// for custom godwoken-scripts contract tests only
+    #[cfg(feature = "unhandled-execution")]
```
keroro520, line 716: Seems a comment "tests of godwoken-scripts use only" is enough, how about omitting the "unhandled-execution" feature?

keroro520 (approved):

jjyr (approved):

PR #823, opened 2022-10-25
chore: cleanup unused code, the total capacity check is handled by the withdrawal::Generator::include_and_verify

jjyr:


blckngm (dismissed):

crates/mem-pool/src/pool.rs
```diff
@@ -1015,20 +1014,6 @@ impl MemPool {
                 continue;
             }
             let capacity: u64 = withdrawal.raw().capacity().unpack();
```
blckngm: Also unused now.

zeroqn (approved):

PR #828, opened 2022-10-25
fix(pool): insert re-injected withdrawals to db

blckngm:
If a block containing withdrawals is reverted, mem pool should insert these withdrawals to db mem-pool withdrawals when re-injecting them. Otherwise producing the next block will fail.

### DIFF
--- a/crates/mem-pool/src/pool.rs
+++ b/crates/mem-pool/src/pool.rs
@@ -1053,7 +1053,21 @@ impl MemPool {
                     let touched_keys = state.tracker_mut().touched_keys().expect("touched keys");
 
                     let withdrawal_hash = withdrawal.hash().into();
-                    db.insert_mem_pool_withdrawal(&withdrawal_hash, withdrawal)?;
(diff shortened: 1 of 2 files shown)


crates/mem-pool/src/pool.rs
```diff
@@ -1062,6 +1063,7 @@ impl MemPool {
                     unused_withdrawals.push(withdrawal_hash);
                 }
             }
+            db.insert_mem_pool_withdrawal(&withdrawal.hash().into(), withdrawal)?;
```
jjyr: We should only re-insert withdrawal in `Ok` branch, in error branch the withdrawal is dropped.

blckngm: Updated.

jjyr (dismissed):

jjyr (dismissed):

jjyr (approved):

Flouse (approved):

PR #832, opened 2022-10-28
feat: add rewind-to-last-valid-block subcommand

blckngm:
Godwoken node operators can use this command to rewind the chain to the last valid block if the node has encountered some bad blocks accidentally due to e.g. version/configuration differences.

Challenge targets for these blocks are also removed, so these blocks can be validated again.

A reference for this command is added to the log message generated when inserting the first bad block.

### DIFF
--- a/crates/chain/src/chain.rs
+++ b/crates/chain/src/chain.rs
@@ -338,7 +338,11 @@ impl Chain {
                         db.rollback()?;
 
                         let block_number = l2block.raw().number().unpack();
-                        log::warn!("bad block #{} found, rollback db", block_number,);
+                        log::warn!(
+                            "bad block #{} found, rollback db. If this is accidental, \
+                            you can rewind bad blocks with the rewind-to-last-valid-block subcommand",
+                            block_number
+                        );
 
                         db.insert_bad_block(&l2block, &global_state)?;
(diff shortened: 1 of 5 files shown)


crates/godwoken-bin/src/main.rs
```diff
@@ -50,32 +53,32 @@ fn generate_example_config<P: AsRef<Path>>(path: P) -> Result<()> {
 // TODO: @zeroqn update clap to v3
 async fn run_cli() -> Result<()> {
     let version = Version::current().to_string();
-    let app = App::new("Godwoken")
+    let app = Command::new("Godwoken")
```
blckngm, line 56: `App` and `SubCommand` are deprecated in clap 3 and removed in clap 4. We should gradually remove usage of deprecated clap 3 features so we can eventually upgrade to clap 4.

crates/chain/src/chain.rs
```diff
@@ -807,6 +811,8 @@ impl Chain {
                             .expect("rewind block should exists");
 
                         db.rewind_block_smt(&block)?;
+                        // XXX: delete the bad block?
```
blckngm, line 814: Bad blocks are deleted in

- https://github.com/godwokenrises/godwoken/pull/821

jjyr (approved):

zeroqn (approved):

PR #835, opened 2022-10-31
Support revert inner state and add syscalls

jjyr:
* Use layered StateDB to operate the state.
  * Implement in memory journals on StateDB, it support reverts nested state before write to RocksDB.
* Support `revert / create` godwoken syscalls.


crates/store/src/state/state_db.rs
```diff
@@ -94,152 +274,306 @@ impl<'a> StateTree<'a> {
...
+        let rev_index = self
+            .revisions
+            .binary_search_by_key(&id, |r| r.id)
+            .map_err(|_id| anyhow!("Invalid revision id"))?;
+        let rev = &self.revisions[rev_index];
+
+        // replay to revert journal
+        let revert_entries = self.journal.revert_entries(rev.journal_len);
+        for entry in revert_entries.into_iter().rev() {
+            entry.revert(self);
```
blckngm: Why not implement snapshot/revert by simply cloning/restoring all the dirty data fields? `im::HashMap` is supposed to be cheap to clone. It would be easier to implement and might be faster too.

jjyr: Yeah, the performance should be samilar. I just pick one from the two choices.

Flouse:
Please fix `cargo` [clippy](https://github.com/godwokenrises/godwoken/actions/runs/3368665467/jobs/5588393202#step:7:789) check

crates/store/src/state/overlay/mem_state.rs
```diff
@@ -90,30 +81,42 @@ impl<'a> State for MemStateTree<'a> {
...
 
     fn get_script(&self, script_hash: &H256) -> Option<packed::Script> {
         self.db()
+            .inner_store()
             .get(COLUMN_SCRIPT, script_hash.as_slice())
-            .or_else(|| self.db().get(COLUMN_SCRIPT, script_hash.as_slice()))
+            .or_else(|| {
+                self.db()
+                    .inner_store()
+                    .get(COLUMN_SCRIPT, script_hash.as_slice())
```
blckngm: Why do we need to get twice?

crates/store/src/state/state_db.rs
```diff
@@ -94,152 +291,342 @@ impl<'a> StateTree<'a> {
...
 
-impl<'a> CodeStore for StateTree<'a> {
+impl<S: CodeStore> CodeStore for StateDB<S> {
     fn insert_script(&mut self, script_hash: H256, script: packed::Script) {
-        self.db()
-            .insert_raw(COLUMN_SCRIPT, script_hash.as_slice(), script.as_slice())
-            .expect("insert script");
+        self.journal.push(JournalEntry::InsertScript {
+            script_hash,
+            prev_exist: self.get_script(&script_hash).is_some(),
```
blckngm: Would `self.dirty_scripts.contains(&script_hash)` be better?

Otherwise if the script exists in self.state, but not in self.dirty_scripts, after revert it would still be in dirty scripts.

crates/store/src/state/state_db.rs
```diff
@@ -94,152 +291,342 @@ impl<'a> StateTree<'a> {
...
+        }
+        // kv
+        for key in state_a.dirty_state.keys().chain(state_b.dirty_state.keys()) {
+            if state_a.get_raw(key) != state_b.get_raw(key) {
+                return false;
+            }
+        }
+
+        true
+    }
```
blckngm, line 562: This method can be implemented much simpler leveraging the `Eq` trait impl for `HashMap`, e.g.:

```rust
    fn cmp_dirty_state(state_a: &MemStateDB, state_b: &MemStateDB) -> bool {
        (
            &state_a.dirty_account_count,
            &state_a.dirty_data,
            &state_a.dirty_scripts,
            &state_a.dirty_state,
        ) == (
            &state_b.dirty_account_count,
            &state_b.dirty_data,
            &state_b.dirty_scripts,
            &state_b.dirty_state,
        )
    }
```

The name could also be a bit more specific, i.e. `dirty_state_eq`.

crates/store/src/state/overlay/mem_state.rs
jjyr: Fixed

crates/store/src/state/state_db.rs
jjyr: For example:

state_a:
- A -> 1 (mem)
- B -> 2 (mem)

state_b:

- A -> 1 (mem)
- B -> 2 (db)

If we compare dirty state it will return false, but this function expect the return is true.

crates/store/src/state/state_db.rs
```diff
@@ -1,60 +1,257 @@
...
+    /// next_revision_id
+    next_revision_id: usize,
+    /// revisions
+    revisions: im::Vector<Revision>,
+    /// dirty state in memory
+    dirty_state: im::HashMap<H256, H256>,
+    /// dirty account count
+    dirty_account_count: Option<u32>,
+    /// dirty scripts
+    dirty_scripts: im::HashMap<H256, packed::Script>,
```
blckngm: I'm wondering whether we should use regular `HashMap`/`Vec` instead of immutable ones if we are using journals. Do we really need fast cloning of a dirty `StateDB`?

blckngm (dismissed):

zeroqn (dismissed):

blckngm (dismissed):

crates/store/src/state/state_db.rs
```diff
@@ -1,60 +1,267 @@
...
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.entries.is_empty()
+    }
+
+    /// Revert journal to len
+    /// Return reverted entries sorted in journal emit order
+    pub fn revert_entries(&mut self, len: usize) -> Vec<JournalEntry> {
+        self.entries.split_off(len)
```
blckngm: Is it possible to use `drain(len..).rev()` instead? That doesn't need to allocate and copy all the reverted entries.

blckngm: I tried it and the borrow checker is not happy. But it can be resolved if the `JournalEntry::revert` method is inlined.

blckngm (approved):

zeroqn (approved):

PR #844, opened 2022-11-07
chore: remove unused crates

jjyr:
* Remove `eoa-mapping` and `ckb-hardfork`.
* Move `gwos/tests` into `crates/tests/script-tests`.


blckngm (dismissed):

blckngm (dismissed):
What about scripts fetched to `.tmp/binaries`?

jjyr:
> What about scripts fetched to `.tmp/binaries`?

I want to remove the binary fetch script and implement a build & cache mechanism. CI refreshes the cache when `/gwos` changes, otherwise it downloads script binaries from the cache.

----

This should be implemented in another PR.

zeroqn (approved):

PR #852, opened 2022-11-11
feat: support increase l2 tx max cycles

jjyr:
1. add fork config
2. fetch consensus parameters from fork config

---------------

Configuration changes:

1. move `backend_switches` to `fork.backend_forks`
2. move `debug_backend_switches` to `debug_backend_forks`

### DIFF
--- a/crates/generator/src/generator.rs
+++ b/crates/generator/src/generator.rs
@@ -3,7 +3,6 @@ use std::{collections::HashSet, sync::Arc, time::Instant};
 use crate::{
     account_lock_manage::AccountLockManage,
     backend_manage::BackendManage,
-    constants::{L2TX_MAX_CYCLES, MAX_READ_DATA_BYTES_LIMIT, MAX_WRITE_DATA_BYTES_LIMIT},
     error::{BlockError, TransactionValidateError, WithdrawalError},
     syscalls::RunContext,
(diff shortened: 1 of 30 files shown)


crates/generator/src/generator.rs
```diff
@@ -734,6 +742,7 @@ impl Generator {
         let backend = self
             .load_backend(block_info.number().unpack(), state, &script_hash)
```
blckngm, line 721: Why not load `max_cycles` here?

jjyr: Debug API support pass `max_cycles` from request. It helps us debug cycles error.

We can remove the argument if we can ensure a polyjuice tx can be 100% verified within the default cycles.

blckngm: How about adding max cycles config to debug backend config?

blckngm: OK. I see that debug max cycles is from RPC params.

So how about taking `override_max_cycles: Option<u64>`, and load from config if it's `None`.

jjyr: Good idea

keroro520 (dismissed):

Flouse (commented):
/itest

Flouse (approved):

blckngm (approved):

PR #858, opened 2022-11-14
feat: determine global state version according to fork height

keroro520:


blckngm (dismissed):
LGTM.

crates/config/src/fork_config.rs
```diff
@@ -43,9 +43,24 @@ pub struct ForkConfig {
     /// Enable this to increase l2 tx cycles limit to 500M
     pub increase_max_l2_tx_cycles_to_500m: Option<u64>,
     pub backend_forks: Vec<BackendForkConfig>,
+
+    /// Bump GlobalState.version from v1 to v2.
+    /// Fork changes:
+    ///   - Optimize finality mechanism
+    ///   - Remove `state_checkpoints` from RawL2Block
+    pub v2: Option<u64>,
```
jjyr: `upgrade_global_state_version_to_v2: Option<u64>`

The version of Godwoken is still v1. The `global_state.version` represents the on-chain scripts verification version, which is different from Godwoken's main version. So we can use a more specific name to avoid confusion.

We should move the field to upper `backend_forks` to keep toml serialization work.

jjyr (changes requested):

crates/config/src/fork_config.rs
keroro520: Make sense!
Updated.

jjyr (approved):

blckngm (approved):

PR #863, opened 2022-11-17
feat(mem-pool): auto-create accounts for native pCKB transfer recipients

zeroqn:
Filter ckb transfer log and save new recipient addresses to mem block. Create accounts after mem pool reset.

require minimal balance: 1 ckb

### DIFF
--- a/crates/mem-pool/src/account_creator.rs
+++ b/crates/mem-pool/src/account_creator.rs
@@ -1,0 +1,218 @@
+use anyhow::{anyhow, bail, Result};
+
+use gw_common::{
+    builtins::{CKB_SUDT_ACCOUNT_ID, ETH_REGISTRY_ACCOUNT_ID, RESERVED_ACCOUNT_ID},
+    ckb_decimal::CKBCapacity,
+    registry_address::RegistryAddress,
+    state::State,
+    H256,
+};
+use gw_generator::account_lock_manage::secp256k1::Secp256k1Eth;
+use gw_types::{
+    core::{AllowedEoaType, ScriptHashType},
+    packed::{
+        BatchCreateEthAccounts, Fee, L2Transaction, LogItemReader, MetaContractArgs,
+        RawL2Transaction, Script, ScriptVec,
+    },
+    prelude::{Builder, Entity, Pack, Reader, Unpack},
+    U256,
+};
+use gw_utils::{
+    script_log::{parse_log, GwLog, GW_LOG_SUDT_TRANSFER},
+    wallet::Wallet,
+    RollupContext,
+};
+use tracing::instrument;
+
+const META_CONTRACT_ACCOUNT_ID: u32 = RESERVED_ACCOUNT_ID;
+const ONE_CKB: u64 = 10u64.pow(8);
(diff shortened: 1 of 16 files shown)


crates/mem-pool/src/account_creator.rs
```diff
@@ -0,0 +1,218 @@
...
+    wallet::Wallet,
+    RollupContext,
+};
+use tracing::instrument;
+
+const META_CONTRACT_ACCOUNT_ID: u32 = RESERVED_ACCOUNT_ID;
+const ONE_CKB: u64 = 10u64.pow(8);
+pub const MIN_BALANCE: u64 = ONE_CKB;
+
+pub struct CkbTransfer;
```
magicalne: For me, it's not straightforward to understand this. It's weird to see the `CkbTransfer` struct, which is empty, in  account_creteor.rs.
How about `fn filter_new_address`  only or `filter_new_created_address`?

crates/mem-pool/src/account_creator.rs
```diff
@@ -0,0 +1,218 @@
...
+
+impl CkbTransfer {
+    pub fn filter_new_address<'a>(
+        logs: impl IntoIterator<Item = LogItemReader<'a>>,
+        state: &impl State,
+    ) -> Option<Vec<RegistryAddress>> {
+        let do_filter = |log: LogItemReader<'_>| -> Result<Option<RegistryAddress>> {
+            if !matches!(log.service_flag().into(), GW_LOG_SUDT_TRANSFER)
+                || !matches!(log.account_id().unpack(), CKB_SUDT_ACCOUNT_ID)
```
blckngm: Question: could this kind of logs be forged?

And nit: no need to use `matches!` when you can use `!=`:

```
            if (log.service_flag().into(), log.account_id().unpack())
                != (GW_LOG_SUDT_TRANSFER, CKB_SUDT_ACCOUNT_ID)
```

crates/mem-pool/src/mem_block.rs
```diff
@@ -93,6 +96,7 @@ impl MemBlock {
         let content = MemBlockContent {
             txs: self.txs.clone(),
             withdrawals: self.withdrawals.clone(),
+            new_addresses: self.new_addresses.drain().collect(),
```
blckngm: Nit: use `std::mem::take` to take `txs`, `withdrawals` and `new_addresses`.

blckngm (dismissed):

crates/mem-pool/src/account_creator.rs
zeroqn: Fixed

crates/mem-pool/src/account_creator.rs
zeroqn: Answer: No. Only sudt contract can emit this log.

Fixed

crates/mem-pool/src/mem_block.rs
zeroqn: Fixed

magicalne (approved):

jjyr (approved):

PR #865, opened 2022-11-18
feat: optimized trace and metrics

zeroqn:
- new gw-telemetry crate for trace and metric
- new chain, block producer, rpc server metrics
- optimize trace instrument
- write trace and span id to log


crates/otel/src/metric.rs
```diff
@@ -0,0 +1,37 @@
...
+pub use prometheus_client::{encoding, metrics::*, registry};
+
+use encoding::text::SendSyncEncodeMetric;
+use registry::Registry;
+
+/// Global metrics registry.
+pub static REGISTRY: Lazy<RwLock<Registry<Box<dyn SendSyncEncodeMetric>>>> =
+    Lazy::new(|| Registry::with_prefix("gw").into());
+
+pub fn global<'a>() -> RwLockWriteGuard<'a, Registry<Box<dyn SendSyncEncodeMetric>>> {
```
blckngm: I think `global_registry` or `registry` would be a better name for this function.

crates/otel/src/metric.rs
```diff
@@ -0,0 +1,37 @@
...
+//! related metrics and they usually change together, it SHOULD live in a
+//! separate metrics module. See the metrics module in gw-chain for an example.
+//!
+//! When you add/modify some metrics, make sure to update the metrics document
+//! in docs/metrics.md.
+//!
+use std::sync::{RwLock, RwLockWriteGuard};
+
+pub use once_cell::sync::{Lazy, OnceCell};
+pub use prometheus_client;
```
blckngm: Why is this module in `otel`? It could be a bit confusing because opentelemetry metrics and openmetrics (prometheus) are two different metrics solution, and we are actually using the latter, not opentelemetry metrics.

crates/rpc-server/src/metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+        registry.register(
+            "execute_transactions",
+            "Number of execute_transaction requests",
+            Box::new(self.execute_transactions.clone()),
+        );
+        registry.register(
+            "in_queue_requests",
+            "Number of in queue requests",
+            Box::new(self.in_queue_requests.clone()),
+        );
```
blckngm: Read-only nodes don't have this metric.

crates/block-producer/src/metrics.rs
```diff
@@ -0,0 +1,205 @@
...
+        registry.register(
+            "local_blocks",
+            "Number of local blocks",
+            Box::new(self.local_blocks.clone()),
+        );
+        registry.register(
+            "submitted_blocks",
+            "Number of submitted blocks",
+            Box::new(self.submitted_blocks.clone()),
+        );
```
blckngm: I still think the original approach (inline, fine grained metrics that lives close to the code that updates it) is superior.

Now metrics for read-only node and full-node are all mixed up, and the metrics doc is no longer accurate. Maybe we should remove it if the effort required to maintain it is not really justified.

Our crates are not exactly perfectly organized, e.g. the block sync client shouldn't really live in the block_producer crate, and the in-queue-request map and the request submitter maybe shouldn't live in the rpc server crate. Defining all metrics in crate level `metrics.rs` module results in weaker cohesion and undesirable coupling - if we want to move e.g. the block sync client to another crate, there will be a lot of churn dealing with its metrics.

crates/rpc-server/src/metrics.rs
zeroqn: Yes. We won't see them in prometheus, because we don't initialize this metric for readonly node.

crates/otel/src/metric.rs
zeroqn: Any suggestion? I want them in one crate. I also consider switching to opentelemetry metric in the future.

crates/block-producer/src/metrics.rs
zeroqn: > I still think the original approach (inline, fine grained metrics that lives close to the code that updates it) is superior.

Original approach introduces at least four different code styles to integrate metrics.
- global structure
- use prometheus client type directly in business logic structure
- pass metric structure all the way down
- register new metric anywhere

To find out where this metric comes from is difficult.  Update metrics for different ops is difficult.

Now
- crate metrics.rs and global structure

> Now metrics for read-only node and full-node are all mixed up, and the metrics doc is no longer accurate. Maybe we should remove it if the effort required to maintain it is not really justified.

Yes, should remove it. 
 
> Our crates are not exactly perfectly organized, e.g. the block sync client shouldn't really live in the block_producer crate, and the in-queue-request map and the request submitter maybe shouldn't live in the rpc server crate. Defining all metrics in crate level `metrics.rs` module results in weaker cohesion and undesirable coupling - if we want to move e.g. the block sync client to another crate, there will be a lot of churn dealing with its metrics.

That why my original plan is define all metrics in gw-metrics crate. Don't bother about metrics if we only want to move code into another crate. Everything still works.

crates/otel/src/metric.rs
blckngm: Why should they be in one crate?

crates/block-producer/src/metrics.rs
blckngm: > > I still think the original approach (inline, fine grained metrics that lives close to the code that updates it) is superior.
> 
> Original approach introduces at least four different code styles to integrate metrics.

It's one style documented in `gw-metrics`:

```
//! Each crate/module/component can define their own metrics and register them
//! to the global `REGISTRY`. To avoid naming conflict, each of them SHOULD use
//! a unique prefix, e.g. the crate name or component name.
//!
//! If it makes sense to define a metrics struct, e.g. when there are a few
//! related metrics and they usually change together, it SHOULD live in a
//! separate metrics module. See the metrics module in gw-chain for an example.
```

>     * global structure

There are actually no global metrics structure after I moved `ChainMetrics` to `gw-chain`.

>     * use prometheus client type directly in business logic structure

It's recommended by the original style. By having a single source of truth, it's much less likely that metrics are not updated when business logic changes.

>     * pass metric structure all the way down

It's the context structure that's passed down, not a metrics structure. The metrics naturally belongs to the context.

>     * register new metric anywhere

It's recommended by the original style.

> To find out where this metric comes from is difficult. Update metrics for different ops is difficult.

Not really. Just search for the metrics name or description.

And why is updating metrics for different ops difficult?

> Now
> 
>     * crate metrics.rs and global structure
> 
> 
> > Now metrics for read-only node and full-node are all mixed up, and the metrics doc is no longer accurate. Maybe we should remove it if the effort required to maintain it is not really justified.
> 
> Yes, should remove it.
> 
> > Our crates are not exactly perfectly organized, e.g. the block sync client shouldn't really live in the block_producer crate, and the in-queue-request map and the request submitter maybe shouldn't live in the rpc server crate. Defining all metrics in crate level `metrics.rs` module results in weaker cohesion and undesirable coupling - if we want to move e.g. the block sync client to another crate, there will be a lot of churn dealing with its metrics.
> 
> That why my original plan is define all metrics in gw-metrics crate. Don't bother about metrics if we only want to move code into another crate. Everything still works.

What I had in mind is that we can have a metrics module under the e.g. `in_queue_request_map` module, if we have to define metrics in a metrics module. Not directly under the rpc-server crate root.

crates/rpc-server/src/metrics.rs
blckngm: Are you sure? Isn't this always registered when `RPC_METRICS` is initialized?

crates/otel/src/metric.rs
zeroqn: Original `gw-metrics` merely expose `REGISTRY`, i think we can merge them into one.

crates/rpc-server/src/metrics.rs
zeroqn: They are "registered" but never create. They won't show up in prometheus side.

crates/block-producer/src/metrics.rs
zeroqn: I create a discussion, we can continue there.

crates/rpc-server/src/metrics.rs
blckngm: What does it mean that they are never “create”?

I just tested it and if a metrics is registered, it will be scraped (by the `scrape` function).

crates/otel/src/metric.rs
blckngm: I think we can rename this crate to `gw-telemetry`. It's more general than otel (opentelemetry) and also easier to understand than an abbreviation.

crates/otel/src/metric.rs
zeroqn: Fixed

crates/rpc-server/src/metrics.rs
zeroqn: I mean [get_or_create](https://docs.rs/prometheus-client/0.18.1/prometheus_client/metrics/family/struct.Family.html#method.get_or_create), they are not queryable.

crates/otel/src/metric.rs
zeroqn: fixed

crates/rpc-server/src/metrics.rs
blckngm: OK. I see that `in_queue_requests` is now a `Family`.

crates/generator/src/generator.rs
```diff
@@ -842,6 +868,20 @@ impl Generator {
...
+        // Record run result
+        {
+            let span = tracing::Span::current();
+            span.record("exit_code", r.exit_code);
+            if r.exit_code != 0 {
+                tracing::error!(error = "non zero exit code")
+            }
+            span.record("execution_cycles", r.cycles.execution);
+            span.record("virtual_cycles", r.cycles.r#virtual);
+            span.record("read_kv_count", r.read_data_hashes.len());
```
jjyr: `r.read_data_hashes` is not kv, please get kv from StateTracker

zeroqn: Fixed. 
Rename "[read|write]_kv_count" to "[read|write]_data_count".
Add "touched_key_count" from StateTracker.

jjyr (approved):

blckngm (approved):

PR #869, opened 2022-11-22
feat: (optionally) support gasless transactions

blckngm:
Support gasless transactions when the `gasless_tx_support` config is present:

```toml
[gasless_tx_support]
entrypoint_address = "entrypoint address in 0x-prefixed hex format"
```

* Sort gasless transactions according to actual transaction gas limit/gas price defined in transaction payload (`UserOperation`).

  (This only affects a full node. Read-only nodes will allow any received gasless transactions, without verifying the entrypoint address.)

Also updated some ethereum related deps.

### DIFF
--- a/crates/config/src/config.rs
+++ b/crates/config/src/config.rs
@@ -54,6 +54,9 @@ pub struct Config {
     pub p2p_network_config: Option<P2PNetworkConfig>,
     #[serde(default)]
     pub sync_server: SyncServerConfig,
+    /// Gasless tx support is enabled when this config presents.
+    #[serde(default)]
+    pub gasless_tx_support: Option<GaslessTxSupportConfig>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
@@ -275,6 +278,12 @@ impl Default for DebugConfig {
             enable_debug_rpc: false,
         }
     }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default)]
+pub struct GaslessTxSupportConfig {
+    /// Gasless tx entrypoint address.
+    pub entrypoint_address: H160,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
--- a/crates/config/src/config.rs
+++ b/crates/config/src/config.rs
@@ -1,4 +1,5 @@
 use ckb_fixed_hash::{H160, H256};
+pub use gw_jsonrpc_types::godwoken::GaslessTxSupportConfig;
 use gw_jsonrpc_types::{
     blockchain::{CellDep, Script},
     ckb_jsonrpc_types::JsonBytes,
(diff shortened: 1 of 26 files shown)


crates/config/src/config.rs
```diff
@@ -54,6 +54,9 @@ pub struct Config {
     pub p2p_network_config: Option<P2PNetworkConfig>,
     #[serde(default)]
     pub sync_server: SyncServerConfig,
+    /// Gasless tx support is enabled when this config presents.
+    #[serde(default)]
+    pub gasless_tx_support: Option<GaslessTxSupportConfig>,
```
RetricSu, line 60: can we expose such config via RPC method `gw_get_node_info` https://github.com/godwokenrises/godwoken/blob/develop/crates/rpc-server/src/registry.rs#L1996 , I want to remove the duplicated config in godwoken-web3

blckngm: OK. Shouldn't be very difficult. ETA this afternoon.

crates/mem-pool/src/fee/types.rs
```diff
@@ -236,15 +240,24 @@ fn parse_l2tx_fee_rate(
...
+                PolyjuiceParser::from_raw_l2_tx(raw_l2tx).context("parse polyjuice args")?;
             // Note: Polyjuice use CKB_SUDT to pay fee by default
-            let poly_args = raw_l2tx_args.as_ref();
-            let gas_limit = u64::from_le_bytes(poly_args[8..16].try_into()?);
-            let gas_price = u128::from_le_bytes(poly_args[16..32].try_into()?);
+            let (gas_limit, gas_price) = if poly_args.gas_price() > 0 {
+                (poly_args.gas(), poly_args.gas_price())
+            } else {
+                // Gasless tx.
+                rollup_context.check_gasless_tx_to_address(
```
jjyr: IMO we shouldn't reject `gas_price=0` txs here, some networks may accept normal txs which `gas_price=0`.

jjyr: We should return `0` as the gas_price if it is, and only return a gas-less version price when to_address is the entry_point contract.

blckngm: Done.

jjyr (dismissed):

crates/config/src/config.rs
blckngm: Done.

crates/rpc-server/src/registry.rs
```diff
@@ -178,6 +178,7 @@ pub struct RegistryArgs<T> {
     pub server_config: RPCServerConfig,
     pub chain_config: ChainConfig,
     pub consensus_config: ConsensusConfig,
+    pub gasless_tx_support_config: Option<GaslessTxSupportConfig>,
```
blckngm, line 181: Passing this down is getting so cumbersome that I am missing DI frameworks😂.

Why don't we just pass the whole config?

RetricSu: As far as I can recall,  historical reason? the RPC `gw_get_node_info` is very simple at the early stage.

jjyr:
PLS fix the CI

jjyr (approved):

zeroqn (approved):

PR #879, opened 2022-11-28
fix(CI): polyjuice integration test

### LINKED ISSUE #875: GWOS EVM MainCI Error
image
https://github.com/godwokenrises/godwoken/actions/runs/3553903432/workflow#L66

magicalne:
Looks like `defaults` cannot be used with `uses` at the same time.
Since the integration test doesn't need a working directory, we can simply delete the default entry. I have tested this in my repo. See more here: https://github.com/magicalne/godwoken/actions/runs/3561861834/jobs/5983131222


magicalne:
Testing EVM main CI.

jjyr (approved):

PR #882, opened 2022-11-29
feat: support non-x86 build

jjyr:
1. Upgrade to `ckb-vm 0.22.0`.
2. Use a separate crate `ckb-vm-aot` to enable the AOT feature on x86.
3. Use the trace machine in non-x86 environments.
4. Copy files from https://github.com/nervosnetwork/ckb-sdk-rust - We met a lot of troubles when upgrading the `ckb-sdk` crate. It contains too many dependencies that may conflict with Godwoken's dependencies, and we want to make the dependencies minimal, so we copy what we need from the repo.


blckngm:
Do we want to check this in CI? Don't know how ready the github hosted M1 runners are, but we can always `cross check` for aarch64-unknown-linux-gnu.

jjyr:
> Do we want to check this in CI? Don't know how ready the github hosted M1 runners are, but we can always `cross check` for aarch64-unknown-linux-gnu.

I will make the change of CI in another PR. It is painful to get it run correctly...

keroro520 (dismissed):
LGTM

crates/tools/Cargo.toml
```diff
@@ -42,4 +43,14 @@ hex = "0.4"
...
+bech32 = "0.8.1"
 tracing-subscriber = { version = "0.3.11", default-features = false, features = ["tracing-log"] }
+parking_lot = "0.12"
+lru = "0.8"
+dashmap = "5.4"
+
+[dev-dependencies]
+ckb-chain-spec = "0.105.1"
+ckb-script = "0.105.1"
+httpmock = "0.6"
```
blckngm: This pulls in a lot of weight...

jjyr: Actually, we can remove the `httpmock`, it only used in the `ckb-sdk` tests...

crates/rpc-client/src/ckb_client.rs
```diff
@@ -97,12 +100,17 @@ impl CKBClient {
     #[instrument(skip_all, fields(tx_hash = %tx_hash.pack()))]
     pub async fn get_transaction(&self, tx_hash: H256) -> Result<Option<Transaction>> {
         let tx_with_status = self.get_transaction_with_status(tx_hash).await?;
-        Ok(tx_with_status
+        tx_with_status
             .and_then(|tx_with_status| tx_with_status.transaction)
             .map(|tv| {
+                let tv: ckb_jsonrpc_types::TransactionView = match tv.inner {
+                    Either::Left(v) => v,
+                    Either::Right(v) => serde_json::from_slice(v.as_bytes())?,
```
blckngm: The bytes are not JSON.

But we won't get a bytes response. We can just use `unreachable!()` here.

jjyr: fixed

blckngm (dismissed):

keroro520 (dismissed):

PR #836, opened 2022-11-01
feat: optimize Godwoken finality mechanism

keroro520:
- https://github.com/godwokenrises/godwoken-scripts/pull/132

PR #883, opened 2022-11-30
feat: deprecate verifications for state_checkpoint_list and prev_state_checkpoint

keroro520:
Currently, Godwoken supports single-round interactive challenge mechanism, in which challengers replay disputed transactions on L1 to detect invalid assertions. We need state root lists for each L2 transaction in order to determine who wins. The state root lists are represented by [`RawL2Block.state_checkpoint_list`](https://github.com/nervosnetwork/godwoken-scripts/blob/b50b8fce5e1a1a2742f2edb6538da5df62b60029/c/godwoken.mol#L107) ~~, and [`RawL2Block.submit_transactions.prev_state_checkpoint`](https://github.com/nervosnetwork/godwoken-scripts/blob/b50b8fce5e1a1a2742f2edb6538da5df62b60029/c/godwoken.mol#L89)~~

Godwoken plans to switch to multiple-round interactive challenge mechanism. So that the state root lists are useless any more.

~~This PR deprecates verifications for `state_checkpoint_list` and `prev_state_checkpoint` but keeps the fields.~~
This PR adds a fork trigger `enforce_correctness_of_state_checkpoint_list(version: u8)`, represents whether check the `state_checkpoint_list` or not. This fork feature is enabled for versions less then v2.

### DIFF
--- a/crates/generator/src/generator.rs
+++ b/crates/generator/src/generator.rs
@@ -465,23 +465,29 @@ impl Generator {
                 Ok(receipt) => receipt,
                 Err(err) => return ApplyBlockResult::Error(err.into()),
             };
-            let expected_checkpoint = state
-                .calculate_state_checkpoint()
-                .expect("calculate_state_checkpoint");
-            let block_checkpoint: H256 = match state_checkpoint_list.get(wth_idx) {
-                Some(checkpoint) => *checkpoint,
-                None => {
-                    return ApplyBlockResult::Error(
-                        BlockError::CheckpointNotFound { index: wth_idx }.into(),
-                    );
-                }
-            };
-            // since the state-validator script will verify withdrawals, we should always pass this check
-            assert_eq!(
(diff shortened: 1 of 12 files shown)


jjyr:
The `GlobalState#version` should control this feature. Since #836 is merged, we should enable this change only if `global_state.version >= 2`

crates/generator/src/generator.rs
```diff
@@ -651,22 +594,6 @@ impl Generator {
             }
         }
 
-        // check post state
-        if !skip_checkpoint_check {
-            let post_merkle_root: H256 = raw_block.post_account().merkle_root().unpack();
```
zeroqn: IMO, we should keep block post state check. Only remove checkpoint list check.

jjyr (approved):

blckngm (approved):

PR #892, opened 2022-12-07
fix(withdrawal): finalized withdrawal take longer time to unlock

zeroqn:
Rollup cell queried from ckb-indexer may be outdated, because PSC store latest rollup cell outpoint in local cells manager. In order to reduce chance of getting outdated outpoint, we query from local cells manager first.

### DIFF
--- a/crates/block-producer/src/psc.rs
+++ b/crates/block-producer/src/psc.rs
@@ -75,8 +75,6 @@ pub struct PSCContext {
     pub chain: Arc<Mutex<Chain>>,
     pub mem_pool: Arc<Mutex<MemPool>>,
     pub block_producer: BlockProducer,
-    // Use mutex to make rust happy. Actually we won't refresh or access this at
-    // the same time.
     pub local_cells_manager: Arc<Mutex<LocalCellsManager>>,
     pub chain_updater: ChainUpdater,
     pub rollup_type_script: Script,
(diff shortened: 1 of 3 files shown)


crates/block-producer/src/psc.rs
```diff
@@ -77,7 +77,7 @@ pub struct PSCContext {
     pub block_producer: BlockProducer,
     // Use mutex to make rust happy. Actually we won't refresh or access this at
     // the same time.
```
blckngm: This comment should be removed (or updated).

crates/block-producer/src/withdrawal_unlocker.rs
```diff
@@ -256,7 +264,8 @@ impl BuildUnlockWithdrawalToOwner for DefaultUnlocker {
     }
 
     async fn query_rollup_cell(&self) -> Result<Option<CellInfo>> {
-        self.rpc_client.query_rollup_cell().await
+        let local_cells_manager = self.local_cells_manager.lock().await;
+        query_rollup_cell(&local_cells_manager, &self.rpc_client).await
```
keroro520, line 268: Because PSC's l1transactions may or may not be sent to CKB, it seems that the withdrawal unlocker may reference a not-sent out-point?

zeroqn: Unlocker will retry later

crates/block-producer/src/psc.rs
zeroqn: Fixed

keroro520 (approved):

jjyr (approved):

PR #901, opened 2022-12-12
fix: check merkle state before process a block

jjyr:


crates/generator/src/generator.rs
```diff
@@ -416,6 +416,22 @@ impl Generator {
             }
         };
 
+        // check prev state
+        if !skip_checkpoint_check {
```
blckngm, line 423: No such variable in scope?

zeroqn (approved):

PR #905, opened 2022-12-13
refactor: move `gw-types` and `gw-common` to `gwos` folder

jjyr:
1. Move `gw-types` and `gw-common` to the `gwos` folder. Then we can directly referent them in the contract by the path.
2. Move smt from `gw-common` to a standalone crate `gw-smt`, so we can simplify the dependency of rust contracts.
3. Upgrade the `capsule` toolchain to `nightly-2021-08-16`
4. Remove `gwos/contracts/tron-account-lock`. Tron lock is not enabled both in the testnet and mainnet.
5. Rename `gw_utils::block_timepoint` to `gw_utils::finalized_timepoint`


crates/block-producer/src/utils.rs
```diff
@@ -20,9 +20,6 @@ pub async fn dump_transaction<P: AsRef<Path>>(dir: P, rpc_client: &RPCClient, tx
 pub fn global_state_last_finalized_timepoint_to_since(global_state: &GlobalState) -> u64 {
     match Timepoint::from_full_value(global_state.last_finalized_timepoint().unpack()) {
         Timepoint::BlockNumber(_) => 0,
-        Timepoint::Timestamp(time_ms) => {
-            // ATTENTION: I am not sure if I am do right. Please review intensively.
-            Since::new_timestamp_seconds(time_ms.saturating_div(1000).saturating_sub(1)).as_u64()
-        }
+        Timepoint::Timestamp(time_ms) => Since::new_timestamp_seconds(time_ms / 1000).as_u64(),
```
blckngm: Is this correct? Maybe we should reuse `greater_since` from psc?

blckngm: If I'm reading the script correctly we want the since timestamp >= global state timestamp. But `t / 1000 * 1000` may be smaller than `t`. So we should use `t / 1000 + 1`.

jjyr: The `withdrawal-lock` contract expects the `withdrawal cell's finalized timestamp <= since`, but you are right.

keroro520:
> Move gw-types and gw-common to the gwos folder

What's the purpose of this change?

jjyr:
> > Move gw-types and gw-common to the gwos folder
> 
> What's the purpose of this change?

By moving these crates into `gwos` folder, scripts can reference these crates by the path.

crates/block-producer/src/block_producer.rs
```diff
@@ -172,11 +172,11 @@ impl BlockProducer {
         let reverted_block_root: H256 = {
             let db = self.store.begin_transaction();
             let smt = db.reverted_block_smt()?;
-            smt.root().to_owned()
+            (*smt.root()).into()
```
blckngm, line 176: Personally I would write `smt.root().clone().into()` as it's more fluent. But it may not be worth changing now.

crates/block-producer/src/block_producer.rs
```diff
@@ -258,17 +258,20 @@ impl BlockProducer {
...
 
                 let keys: Vec<H256> = collected_block_hashes.into_iter().collect();
                 for key in keys.iter() {
                     log::info!("submit revert block {:?}", hex::encode(key.as_slice()));
                 }
-                let proof = block_smt
-                    .merkle_proof(keys.clone())?
-                    .compile(keys.clone())?;
+                let proof = {
+                    let smt_keys: Vec<_> = keys.iter().map(|k| (*k).into()).collect();
```
blckngm: The compiler *should* be able to optimize this to a no-op:

https://stackoverflow.com/questions/48308759/how-do-i-convert-a-vect-to-a-vecu-without-copying-the-vector

Without `repr(transparent)` on SMT's H256, it's not really safe to skip this copying with unsafe code.

blckngm (dismissed):

crates/block-producer/src/block_producer.rs
blckngm: Actually, I just tried it and the optimization is much more likely to kick in when using `into_iter`. Consider using `into_iter` or construct a `Vec<SMTH256>` in the first place.

zeroqn (approved):

blckngm (approved):

PR #940, opened 2023-01-11
feat!: use delegate cell lock as rollup cell lock

blckngm:
Use delegate cell lock to support managing block producer key via e.g. multisig lock.

Delegate cell lock is a simple lock that delegates verification to another lock via a “delegate cell”. Delegate cell can be managed separately by e.g. a multisig lock.


blckngm:
Some gw-tools args have changed, see kicker PR:

- https://github.com/godwokenrises/godwoken-kicker/pull/338

~Edit: it should be still compatible with current kicker~

Edit: no, need to modify `scripts-config.json`

crates/config/Cargo.toml
```diff
@@ -18,6 +18,3 @@ serde = { version = "1.0", features = ["derive"] }
 hex = "0.4"
 lazy_static = "1.4"
 toml = "0.5"
-
-[dev-dependencies]
-gw-utils = { path = "../utils" }
```
blckngm: Changed this because circular dev-dependency breaks rust-analyzer. @jjyr

gwos/contracts/delegate-cell-lock/src/entry.rs
```diff
@@ -0,0 +1,44 @@
...
+    ckb_constants::Source,
+    high_level::{
+        load_cell_data, load_cell_lock_hash, load_cell_type_hash, load_script, load_witness_args,
+        QueryIter,
+    },
+};
+
+use crate::error::Error;
+
+/// Lock args: type hash (and usually type-id) of delegate-cell.
```
jjyr, line 17: We should explain the `delegate-cell` (admin cell) and the `key-cell` (block producer's cell) with an actual lock script to help reviewers understand this script.

* `delegate-cell` plays the role of administrator. We assume it is controlled by the [secp256k1 multisig lock](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0024-ckb-genesis-script-list/0024-ckb-genesis-script-list.md#secp256k1multisig).
* `key-cell` plays the role of the singleton block producer. We assume it is controlled by the [secp256k1 blake160 lock](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0024-ckb-genesis-script-list/0024-ckb-genesis-script-list.md#secp256k1blake160).

Please add the explanation in the comment.

blckngm: Added.

jjyr (dismissed):

gwos/contracts/delegate-cell-lock/src/entry.rs
```diff
@@ -0,0 +1,50 @@
...
+    let script = load_script()?;
+    let args = script.as_reader().args().raw_data();
+
+    let witness_arg = load_witness_args(0, Source::GroupInput)?;
+    if witness_arg.as_reader().lock().is_some() {
+        return Err(Error::InvalidWitnessArgs);
+    }
+
+    // Get expected type hash from args.
+    let expected_type_hash: [u8; 32] = args.try_into().map_err(|_| Error::InvalidArgs)?;
```
XuJiandong, line 32: It might be better when args can be longer than 32 bytes. Better for upgrading.

gwos/contracts/delegate-cell-lock/src/entry.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    // Load cell dep data by type hash.
+    let cell_dep_idx = QueryIter::new(load_cell_type_hash, Source::CellDep)
+        .position(|type_hash| type_hash == Some(expected_type_hash))
+        .ok_or(Error::CellDepNotFound)?;
+    let data = load_cell_data(cell_dep_idx, Source::CellDep)?;
+    let expected_lock_hash_160: [u8; 20] =
```
XuJiandong, line 39: Same as above.

gwos/contracts/delegate-cell-lock/src/entry.rs
```diff
@@ -0,0 +1,50 @@
...
+///
+/// Delegate-cell data should be blake160 hash of a lock script.
+///
+/// Unlock is successful if there is an input with a matching lock script.
+pub fn main() -> Result<(), Error> {
+    let script = load_script()?;
+    let args = script.as_reader().args().raw_data();
+
+    let witness_arg = load_witness_args(0, Source::GroupInput)?;
+    if witness_arg.as_reader().lock().is_some() {
```
XuJiandong, line 27: The witness is not used in script.

XuJiandong (changes requested):

gwos/contracts/delegate-cell-lock/src/entry.rs
blckngm: What is expected to be upgraded here? It's unlikely that the length of “type hash” will change.

gwos/contracts/delegate-cell-lock/src/entry.rs
blckngm: How about requiring this data to:

* has a length >= 20
* and is a prefix of the lock hash

gwos/contracts/delegate-cell-lock/src/entry.rs
blckngm: This is discussed in the internal issue. If we delegate to e.g. secp256k1/blake160, the witnesses of this lock will not be checked by it, so we added some basic sanity checks on the first witness arg.

zeroqn (dismissed):

zeroqn (dismissed):

gwos/contracts/delegate-cell-lock/src/entry.rs
jjyr: IMO it is fair enough to take the first 20 bytes. There is no benefit to accepting a variant length lock hash.

jjyr (dismissed):

jjyr:
@sopium please resolve conflicts

jjyr (approved):

PR #946, opened 2023-01-11
feat: builtin consensus config and binaries

jjyr:
1. Store builtin binaries into https://github.com/godwokenrises/godwoken-builtin-binaries
2. Add a GitHub workflow to push binaries. See [README.md](https://github.com/godwokenrises/godwoken/blob/8928589ed97341e141ec2c065efa7d03688113f9/crates/builtin-binaries/README.md) to learn how to add new binaries.
3. Refactor config, separate consensus configs from node config, and builtin `mainnet` and `testnet` consensus config to reduce DevOps work.
4. Build godwoken built-in binaries into the bin; this increases the `godwoken` bin from `44MB` to `54MB`.

------

Generated example.

``` toml
node_mode = 'readonly'
contract_log_config = 'verbose'

[consensus]
builtin = 'Testnet'

[rpc_client]
ckb_url = ''

[rpc_server]
listen = ''
enable_methods = []

[debug]
output_l1_tx_cycles = true
expected_l1_tx_upper_bound_cycles = 350000000
debug_tx_dump_path = 'debug-tx-dump'
enable_debug_rpc = false

[block_producer]
check_mem_block_before_submit = false
fee_rate = 1000
local_limit = 5
submitted_limit = 5
block_interval_secs = 8

[block_producer.block_producer]
address_type = 'Eth'
address = '0x'
[block_producer.challenger_config.rewards_receiver_lock]
code_hash = '0x0000000000000000000000000000000000000000000000000000000000000000'
hash_type = 'data'
args = '0x'

[mem_pool]
execute_l2tx_max_cycles = 100000000
restore_path = 'mem_block'

[mem_pool.mem_block]
max_deposits = 100
max_withdrawals = 100
max_txs = 1000
deposit_block_timeout = 150
deposit_timestamp_timeout = 1200000
deposit_epoch_timeout = 1
deposit_minimal_blocks = 0
max_cycles_limit = '18446744073709551615'

[mem_pool.extra]
allowed_sudt_proxy_creator_account_id = []
sudt_proxy_code_hashes = []

[store]
path = ''

[sync_server]
buffer_capacity = 16
broadcast_channel_capacity = 1024
```


crates/builtin-binaries/build.rs
```diff
@@ -0,0 +1,104 @@
...
+        (
+            "godwoken-polyjuice-v1.4.5/generator",
+            h256!("0xee766d4a0e56a4c46f8917d52847a20bf9dcc3cd38f705faceec81dc39cdd53a"),
+        ),
+        (
+            "godwoken-polyjuice-v1.5.0/generator",
+            h256!("0xf89d4eb312fa4b8df9a3acc0a25f870201106825d9a609612c61aa8263cda1b8"),
+        ),
+        (
+            "godwoken-polyjuice-v1.5.1/generator",
```
Flouse: I think we should use a different tag for the version fixed by https://github.com/godwokenrises/godwoken/pull/921 
`Polyjuice 1.5.1` is used by https://github.com/godwokenrises/godwoken/tree/1.7.4/gwos-evm

Flouse (changes requested):

crates/builtin-binaries/build.rs
jjyr: Yeah, I will change it to 1.5.2.

.github/workflows/gwos-evm-ci.yml
```diff
@@ -69,3 +69,9 @@ jobs:
         MANUAL_BUILD_POLYJUICE=true
         POLYJUICE_GIT_URL=https://github.com/${{ github.repository }}
         POLYJUICE_GIT_CHECKOUT=${{ github.ref }}
+        MANUAL_BUILD_WEB3=true
+        MANUAL_BUILD_WEB3_INDEXER=true
+        WEB3_GIT_URL="https://github.com/${{ github.repository }}"
+        WEB3_GIT_CHECKOUT=${{ github.ref }}
+        GODWOKEN_KICKER_REPO=jjyr/godwoken-kicker
+        GODWOKEN_KICKER_REF=build-web3-indexer-in-monorepo
```
blckngm: Is this integration test still necessary if we use `MANUAL_BUILD_POLYJUICE=true` in `godwoken-tests.yml`?

crates/config/src/consensus/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+        config: Box<ForkConfig>,
+    },
+}
+
+impl Consensus {
+    pub fn get_config(&self) -> Cow<ForkConfig> {
+        match self {
+            Consensus::Builtin { builtin } => match builtin {
+                BuiltinConsensus::Mainnet => Cow::Owned(builtins::mainnet()),
+                BuiltinConsensus::Testnet => Cow::Owned(builtins::testnet()),
```
blckngm: Is it possible to make `builtins::testnet`/`builtins::mainnet` return `&'static ForkConfig`? Then this function can return `&ForkConfig`.

crates/config/src/consensus/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+pub enum Consensus {
+    /// Builtin consensus config
+    Builtin {
+        /// The identifier of the builtin consensus config.
+        builtin: BuiltinConsensus,
+    },
+    /// Customized config
+    Config {
+        /// The config
+        config: Box<ForkConfig>,
```
blckngm, line 26: Why boxing here?

crates/config/src/consensus/mod.rs
blckngm: Or maybe implement `Deref<Target = ForkConfig>` for `Consensus`.

crates/tools/Cargo.toml
```diff
@@ -15,10 +15,10 @@ anyhow = "1.0"
 thiserror = "1.0"
 toml = "0.5"
 tempfile = "3.1"
-lazy_static = "1.3"
+lazy_static = "1.4"
 secp256k1 = "0.24"
 sha3 = "0.10.6"
-reqwest = { version = "0.11", features = ["json"] }
+reqwest = { version = "0.11", features = ["json", "blocking"] }
```
blckngm, line 21: Is blocking client actually used?

crates/utils/src/checksum.rs
```diff
@@ -0,0 +1,15 @@
+use anyhow::Result;
+use gw_types::h256::H256;
+use sha2::{Digest, Sha256};
+use std::path::Path;
+
+pub fn content_checksum(content: &[u8]) -> Result<H256> {
```
blckngm: This function won't fail, it can return just `H256`.

crates/utils/src/checksum.rs
```diff
@@ -0,0 +1,15 @@
...
+    let mut hasher = Sha256::new();
+    hasher.update(&content);
+    Ok(hasher.finalize().into())
```
blckngm: Nit: just use `Sha256::digest(content).into()`.

crates/builtin-binaries/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+license = "MIT"
+description = "Godwoken builtin binaries"
+repository = "https://github.com/godwokenrises/godwoken"
+include = ["src/**/*", "Cargo.toml", "build.rs", "builtin/**/*", "README.md"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+default = []
+no-builtin = []
```
blckngm, line 15: Cargo features are supposed to be additive, so it might be better to have a `builtin` feature to include builtin scripts rather than having a `no-builtin` feature.

Also, would it be better to default to not having built-in scripts, and only embed built-in scripts on release?

.github/workflows/gwos-evm-ci.yml
jjyr: Nope, we can merge these two actions

crates/config/src/consensus/mod.rs
jjyr: To remove the clips warning large_enum_variant.

crates/tools/Cargo.toml
jjyr: Yes, this feature is enabled in gw-dynamic-config crate, since I removed the crate in this pr, I need to add the feature back to make compiling work.

crates/builtin-binaries/Cargo.toml
jjyr: After a second thought, I think the no builtin is makes sense.

I agreed active style feature is better in some case. But there are no_std feature in the rust. BTW the crate’s name is “builtin binaries”.

Since builtin binaries would be the recommended way to start a Godwoken node, I think we should make it default in the compiling.  Otherwise, developer may find the default way not working. I try to avoid that.

crates/config/src/consensus/mod.rs
blckngm: If we need to box it, maybe use `Arc` to make clone cheap.

crates/config/src/consensus/mod.rs
jjyr: Changed the return type to `&ForkConfig`

crates/config/src/consensus/mod.rs
jjyr: Arc does not implement serde; we need to do it manually. 
We rarely use `ForkConfig#clone` during the node running, so IMO, Box is enough.

crates/builtin-binaries/Cargo.toml
Flouse: Does this mean that we have to build a special version/image for Godwoken internal alphanet?

jjyr: It's an unrelated topic. But yes.

A method to simplify the alphanet config is adding it into the built-in.

Flouse (approved):

blckngm (approved):

.github/workflows/godwoken-tests.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches: [develop, master, ci, 'v1*']
     tags: ["v*.*.*"]
+    paths:
+      - 'gwos-evm/**'
```
blckngm, line 8: Doesn't this mean that this test is only triggered when `gwos-evm` changes? Doesn't feel right.

jjyr: You are right, the paths should be removed

jjyr: Fixed in https://github.com/godwokenrises/godwoken/pull/950/files

PR #960, opened 2023-01-30
refactor(web3): use tokio; remove smol and sentry

jjyr:
* Use `tokio` to replace `smol`.
* Remove `sentry`

### DIFF
--- a/web3/crates/indexer/src/insert_l2_block.rs
+++ b/web3/crates/indexer/src/insert_l2_block.rs
@@ -1,7 +1,9 @@
 use std::{convert::TryFrom, str::FromStr};
 
 use anyhow::Result;
+use futures::StreamExt;
 use gw_types::U256;
+use itertools::Itertools;
 use rust_decimal::Decimal;
 use sqlx::{
     postgres::PgRow,
@@ -14,13 +16,9 @@ use sqlx::{
 use sqlx::{Postgres, QueryBuilder};
 
 use crate::{
-    cpu_count::CPU_COUNT,
     pool::POOL_FOR_UPDATE,
     types::{Block, Log, Transaction, TransactionWithLogs},
 };
-
-use itertools::Itertools;
-use rayon::prelude::*;
 
 extern crate num_cpus;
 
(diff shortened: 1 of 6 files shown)


web3/crates/indexer/src/insert_l2_block.rs
```diff
@@ -408,23 +408,22 @@ pub async fn update_web3_txs_and_logs(
...
-                .collect::<Result<Vec<_>>>()?;
+        let mut futs = Vec::new();
+        for query_builder_vec in &mut logs_query_slice {
+            for query_builder in query_builder_vec {
+                let query = query_builder.build();
+                futs.push(query.execute(&*POOL_FOR_UPDATE));
+            }
+        }
+        for fut in futs {
+            fut.await?;
```
jjyr: The futures are spawned from the `sqlx` crate, so `JoinSet` can't be used here. Do you know how to do it idiomatically? @sopium

blckngm: I would just use stream and `buffer_unordered` instead of slicing:

```rust
        let mut results_stream = futures::stream::iter(logs_querys.iter_mut())
            .map(|query| query.build().execute(&*POOL_FOR_UPDATE))
            .buffer_unordered(POOL_FOR_UPDATE.size() as usize);

        while let Some(r) = results_stream.next().await {
            r?;
        }
```

web3/crates/indexer/src/insert_l2_block.rs
```diff
@@ -178,33 +173,23 @@ pub async fn insert_web3_txs_and_logs(
...
-                .collect();
-            (DbTransaction::try_from(web3_tx_with_logs.tx), db_logs)
-        })
-        .collect::<(Vec<_>, Vec<_>)>();
-    let txs = txs.into_iter().collect::<Result<Vec<_>>>()?;
-    let logs = logs.into_iter().collect::<Result<Vec<_>>>()?;
-    let logs = logs.into_iter().flatten().collect::<Vec<_>>();
+    let mut txs = Vec::with_capacity(web3_tx_with_logs_vec.len());
+    let mut logs = Vec::with_capacity(txs.len() * 2);
+    for (i, web3_tx_with_logs) in web3_tx_with_logs_vec.into_iter().enumerate() {
```
jjyr, line 181: Not a CPU heaven calculation, we can just use iter.

web3/crates/indexer/src/insert_l2_block.rs
```diff
@@ -312,38 +287,22 @@ pub async fn update_web3_txs_and_logs(
...
-    let logs_len = logs.len();
-    let txs_len = txs.len();
-
-    let size = logs_len / 4;
-    let final_size = if size > INSERT_LOGS_BATCH_SIZE || size == 0 {
-        INSERT_LOGS_BATCH_SIZE
-    } else {
-        size
-    };
+    for (i, web3_tx_with_logs) in web3_tx_with_logs_vec.into_iter().enumerate() {
```
jjyr, line 305: Not a CPU heaven calculation, use iter is OK

PR #961, opened 2023-01-31
Fix fuzz test and enable fuzz testing CI

magicalne:
Migrate fuzz testing from https://github.com/godwokenrises/godwoken-polyjuice/pull/181 and enable fuzz testing CI.


gwos-evm/polyjuice-tests/fuzz/gw-syscall-simulator/src/lib.rs
```diff
@@ -0,0 +1,709 @@
...
+    }
+    fn get_data(&self, data_hash: &H256) -> Option<Bytes> {
+        self.state.get_data(data_hash)
+    }
+}
+
+static HOST: Lazy<Mutex<GodwokenHost>> = Lazy::new(|| Mutex::new(GodwokenHost::new()));
+
+#[no_mangle]
+pub extern "C" fn ckb_exit(code: i8) -> i32 {
```
blckngm: Nit: strictly speaking these functions should all be `unsafe`.

blckngm (dismissed):

gwos-evm/polyjuice-tests/fuzz/gw-syscall-simulator/src/lib.rs
blckngm: I mean all public functions that involves raw pointers.

magicalne: So you suggest turning all those functions into `unsafe extern` functions? So people know it is unsafe to use.

BTW, those `fn`s are supposed to be used in C code only, where safety is not the concern, I guess. Especially for the fuzz testing, I hope they crash. 🤪

blckngm: Yeah, I understand that these are supposed to be called only by C code. So it's just a suggestion.

magicalne: Done

blckngm (dismissed):

gwos-evm/polyjuice-tests/fuzz/gw-syscall-simulator/src/lib.rs
```diff
@@ -0,0 +1,714 @@
...
+    hasher.finalize(&mut data_hash);
+    // insert data hash into SMT
+    let data_hash_key = build_data_hash_key(&data_hash);
+    let mut host = HOST.lock().unwrap();
+    host.state
+        .update_raw(data_hash_key, H256::one())
+        .expect("gw store data");
+    host.state.insert_data(data_hash.into(), data);
+
+    host.run_result.write_data_hashes.insert(data_hash_key);
```
jjyr: https://github.com/godwokenrises/godwoken/blob/develop/crates/generator/src/syscalls/mod.rs#L411

`data_hash_key` should not be stored into `write_data_hashes`

gwos-evm/polyjuice-tests/fuzz/gw-syscall-simulator/src/lib.rs
```diff
@@ -0,0 +1,714 @@
...
+        Err(_err) => {
+            return ERROR;
+        }
+    };
+
+    let account_nonce_key = build_account_field_key(id, GW_ACCOUNT_NONCE_TYPE);
+    host.state
+        .update_raw(account_nonce_key, H256::zero())
+        .expect("account nonce key");
+    host.run_result.write_data_hashes.insert(account_nonce_key);
```
jjyr: https://github.com/godwokenrises/godwoken/blob/develop/crates/store/src/state/state_db.rs#L499

The `write_data_hashes` and `read_data_hashes` are only related to these syscalls https://github.com/godwokenrises/godwoken/blob/develop/crates/generator/src/syscalls/mod.rs#L396-L428 we shouldn't insert state key into the `data_hashes`.

Please fix it.

jjyr (changes requested):

gwos-evm/polyjuice-tests/fuzz/gw-syscall-simulator/src/lib.rs
magicalne: Fixed. Those data hashes were stored in the wrong place. And run_result cannot be covered in the fuzz testing, so we can just delete them safely.

gwos-evm/polyjuice-tests/fuzz/gw-syscall-simulator/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+    if let Some(tx) = &HOST.lock().unwrap().tx {
+        store_data(addr, len, 0, tx);
+        SUCCESS
+    } else {
+        GW_ITEM_MISSING
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn gw_load_block_info(addr: *mut c_void, len: *mut u64) -> c_int {
```
Flouse, line 283: Some extern functions do not require unsafe.

> https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html?highlight=no_mangle#calling-rust-functions-from-other-languages

Flouse (approved):
LGTM

gwos-evm/polyjuice-tests/fuzz/gw-syscall-simulator/src/lib.rs
magicalne: @Flouse Here is the context about turning into `unsafe extern fn`.

gwos-evm/polyjuice-tests/fuzz/gw-syscall-simulator/src/lib.rs
blckngm: Being `extern` does not require the functions to be `unsafe`, but many of these functions are actually `unsafe`, i.e. it's possible to cause UB if they are called by safe rust functions with incorrect parameters.

zeroqn (approved):

jjyr (approved):

PR #971, opened 2023-02-05
feat: send pending upgrade l1 tx at certain height

jjyr:
We add a new option, `ForkConfig#pending_l1_upgrades`, which allows the block producer to add an l1 upgrading tx into the config. The l1 upgrading tx will be sent just before the l2 block at the targeting height.


Simon-Tl (dismissed):

crates/block-producer/src/psc.rs
```diff
@@ -736,6 +766,73 @@ async fn median_gte(rpc_client: &RPCClient, timestamp_millis: u64) -> Result<()>
     Ok(())
 }
 
+/// Submit next pending l1 upgrade
+async fn submit_pending_l1_upgrade(ctx: &PSCContext) -> Result<()> {
+    let snap = ctx.store.get_snapshot();
+    let block_number = snap
+        .get_last_confirmed_block_number_hash()
```
zeroqn: `get_last_submitted_block_number_hash()`?

Target `l1_upgrade.height` may be submitted.

jjyr: This function is called just before the `submit_next_block`, so the `height + 1` block hasn't been submitted yet. https://github.com/godwokenrises/godwoken/pull/971/files#diff-b91893c4ad9a6897506ded8d9e15175017791e9aa4f256b093d0e1402da23bc0R261

crates/block-producer/src/psc.rs
```diff
@@ -736,6 +766,73 @@ async fn median_gte(rpc_client: &RPCClient, timestamp_millis: u64) -> Result<()>
...
+        .find(|l1_upgrade| l1_upgrade.height == block_number)
+    {
+        let tx: Transaction = l1_upgrade.signed_transaction.clone().into();
+        // We can't recover from errors when confirming l1 upgrade tx,
+        // both deadcell error and unknwown cell is unacceptable, so we just throw it
+        poll_tx_confirmed(&ctx.rpc_client, &tx).await?;
+        log::info!("l1 upgrade tx confirmed");
+        ctx.local_cells_manager.lock().await.confirm_tx(&tx);
+    }
+    Ok(())
```
Flouse, line 816: Maybe we should reload `ContractsCellDepManager` when the `upgrade transaction` is confirmed, because the ContractsCellDepManager.deps maybe changed by the `upgrade transaction`.

What do you think?

jjyr: Good catch

jjyr: Fixed

crates/block-producer/src/psc.rs
zeroqn: Suppose we want to upgrade at height 100 (`l1_upgrade.height = 100`). 

The code only works as expected when 99 is confirmed but 100 is not submitted. Since we submit 3 blocks in chained txs by default. 100 aka `height + 1` may be submitted.

crates/block-producer/src/psc.rs
```diff
@@ -736,6 +766,79 @@ async fn median_gte(rpc_client: &RPCClient, timestamp_millis: u64) -> Result<()>
...
+
+        // We can't recover from errors when sending l1 upgrade tx, so we just throw it
+        send_transaction_or_check_inputs(&ctx.rpc_client, &tx).await?;
+        log::info!("tx sent");
+
+        // Refresh block producer's contracts deps
+        ctx.block_producer
+            .contracts_dep_manager()
+            .refresh(Some(tx))
+            .await?;
```
blckngm, line 786: Do we really need to add the transaction to local cells manager if we are directly getting deps from it? The upgrade tx shouldn't directly involve any cells used/created by submission txs.

blckngm: (Assuming that the upgrade tx doesn't use the same wallet/lock as the block producer to pay fee.)

crates/block-producer/src/psc.rs
```diff
@@ -252,6 +258,18 @@ async fn run(state: &mut ProduceSubmitConfirm) -> Result<()> {
             let context = state.context.clone();
             submit_handle.replace_with(tokio::spawn(async move {
                 loop {
+                    match submit_pending_l1_upgrade(&context).await {
+                        Ok(()) => (),
+                        Err(err) => {
+                            if err.is::<ShouldResyncError>() || err.is::<ShouldRevertError>() {
```
blckngm: If submitting the upgrade tx fails, resync or revert may not be helpful.

crates/block-producer/src/psc.rs
```diff
@@ -271,6 +289,18 @@ async fn run(state: &mut ProduceSubmitConfirm) -> Result<()> {
             let context = state.context.clone();
             confirm_handle.replace_with(tokio::spawn(async move {
                 loop {
+                    match confirm_pending_l1_upgrade(&context).await {
+                        Ok(()) => (),
+                        Err(err) => {
+                            if err.is::<ShouldResyncError>() || err.is::<ShouldRevertError>() {
```
blckngm: If the upgrade tx cannot be confirmed, resync or revert may not be helpful.

jjyr: IMO throw the err is enough, What are you suggest to do here?

crates/block-producer/src/psc.rs
jjyr: Ok, we should use `get_last_submitted_block_number_hash` here

crates/block-producer/src/psc.rs
jjyr: Removed the local cell manager injection

crates/block-producer/src/psc.rs
jjyr: Fixed

blckngm (approved):

Flouse (approved):

PR #982, opened 2023-02-12
fix(withdrawal_unlocker): remove dry_run_transaction and fix error display

blckngm:


crates/rpc-client/src/utils.rs
```diff
@@ -30,7 +30,7 @@ impl TracingHttpClient {
         self.inner.url()
     }
 
-    #[instrument(target = "gw-rpc-client", skip_all, err, fields(method, params = field::Empty))]
+    #[instrument(target = "gw-rpc-client", skip_all, fields(method, params = field::Empty))]
```
blckngm, line 33: Removed the error from trace because displaying the error without error chain is not very useful.

zeroqn (dismissed):

Flouse (dismissed):

jjyr:
Please merge to `1.12-rc`. We should publish a patch version.

Flouse:
- replaced by #992

PR #983, opened 2023-02-13
feat(gw-tools)!: enhance several commands

blckngm:
* Deploy scripts with chained txs, this is much faster than deploying them one-by-one.

* Don't use ckb-cli. It often has breaking changes. Just use our `TransactionSkeleton` and `Wallet`. I have also added various helper functions to make this easier.

* Enhance generate-config so that kicker doesn't need to modify or read generated config files with sed/awk. See https://github.com/godwokenrises/godwoken-kicker/pull/345.

* Removed unused ckb-sdk code.

(For reviewers: review from the second commit, the first one is the same as #940)


jjyr:
#940 is merged, please rebase this PR

jjyr (approved):

zeroqn (approved):

PR #1011, opened 2023-03-02
fix(polyjuice): native transfer needs check gas limit before pay fee

magicalne:
Native transferring doesn't check the actual gas used. This makes it happens that charging more fee than the gas limit. Or in another sense, the gas limit is ignored.  This occurs when transferring to an unregistered address.
We can fix this by comparing `gas_used` and `gas_limit` before paying the fee.


zeroqn (approved):

Flouse (approved):
I think this change requires a new poly_backend_fork, right?
@magicalne

magicalne:
> I think this change requires a new poly_backend_fork, right?

Yes, this change is not compatible.

jjyr (approved):

Flouse:
> > I think this change requires a new poly_backend_fork, right?
> 
> Yes, this change is not compatible.

cc @Simon-Tl

PR #1020, opened 2023-03-09
feat: trace state changes

blckngm:


jjyr (approved):

magicalne (approved):

crates/store/src/schema.rs
```diff
@@ -37,6 +37,9 @@ pub const COLUMN_ACCOUNT_SMT_LEAF: Col = 10;
 pub const COLUMN_BLOCK_SMT_BRANCH: Col = 11;
 /// Column block SMT leaf
 pub const COLUMN_BLOCK_SMT_LEAF: Col = 12;
+/// Block hash -> block state changes json.
+// TODO: migration to clear this column.
+pub const COLUMN_BLOCK_STATE_CHANGES: Col = 13;
```
Flouse, line 42: What is the plan of this TODO migration?

blckngm: It's a bit complicated to add migrations now because the smt-trie migration may or may not exist, so it's hard to decide on a version for any new migrations...

And if we are only use this on new nodes, which seems to be the case now, we don't really need a migration.

PR #1022, opened 2023-03-13
fix(syscall): handle invalid input in bn256 syscalls

magicalne:
Handle errors in bn256 and return errors for syscalls.

### DIFF
--- a/crates/generator/src/syscalls/bn.rs
+++ b/crates/generator/src/syscalls/bn.rs
@@ -16,20 +16,18 @@ pub enum BnError {
 
 type Result<T> = std::result::Result<T, BnError>;
 fn read_pt(buf: &[u8]) -> Result<G1> {
-    let map_err = |err| BnError::G1FieldError(err);
+    let map_err = BnError::G1FieldError;
     let px = Fq::from_slice(&buf[0..32]).map_err(map_err)?;
     let py = Fq::from_slice(&buf[32..64]).map_err(map_err)?;
     Ok(if px == Fq::zero() && py == Fq::zero() {
         G1::zero()
     } else {
-        AffineG1::new(px, py)
-            .map_err(|err| BnError::GroupError(err))?
-            .into()
+        AffineG1::new(px, py).map_err(BnError::GroupError)?.into()
     })
 }
 
 fn read_fr(buf: &[u8]) -> Result<Fr> {
-    Ok(Fr::from_slice(buf).map_err(|err| BnError::G1FieldError(err))?)
+    Fr::from_slice(buf).map_err(BnError::G1FieldError)
 }
 
 pub fn add(input: &[u8]) -> Result<[u8; 64]> {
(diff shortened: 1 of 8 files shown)


crates/generator/src/syscalls/bn.rs
```diff
@@ -1,27 +1,36 @@
...
-fn read_pt(buf: &[u8]) -> Result<G1, Error> {
-    let px = Fq::from_slice(&buf[0..32]).map_err(|_| Error("invalid pt"))?;
-    let py = Fq::from_slice(&buf[32..64]).map_err(|_| Error("invalid pt"))?;
+type Result<T> = std::result::Result<T, BnError>;
+fn read_pt(buf: &[u8]) -> Result<G1> {
+    let map_err = BnError::G1FieldError;
+    let px = Fq::from_slice(&buf[0..32]).map_err(map_err)?;
+    let py = Fq::from_slice(&buf[32..64]).map_err(map_err)?;
     Ok(if px == Fq::zero() && py == Fq::zero() {
         G1::zero()
```
jjyr, line 23: Is this a breaking change? (is `AffineG1(Fq::zero(), Fq::zero()) == G1::zero()`? )

jjyr (dismissed):

crates/generator/src/syscalls/bn.rs
magicalne: Well, actually I don't break any change here. But after checking the code, looks like  `AffineG1(Fq::zero(), Fq::zero())` and `G1::zero()` are not equivalent.
`G1::zero()` represents:
```rust
        G {
            x: P::Base::zero(),
            y: P::Base::one(),
            z: P::Base::zero(),
        }
```

And for `AffineG1(Fq::zero(), Fq::zero())`:
```rust
      G {
           x: x,
           y: y,
           z: P::Base::one(),
     };
```

@Flouse Do you remember about this? Is that intended?

crates/generator/src/syscalls/mod.rs
```diff
@@ -565,29 +569,39 @@ impl<'a, 'b, S: State + CodeStore + JournalDB, C: ChainView, Mac: SupportMachine
...
-                    log::error!("{}", err_msg);
-                    VMError::Unexpected(err_msg)
-                })?;
-                store_data(machine, output.as_slice())?;
-
-                machine.set_register(A0, Mac::REG::from_u8(SUCCESS));
+                log::debug!("[bn256 input]: {:?}", &input[0..32]);
+                log::debug!("[bn256 input]: {:?}", &input[32..64]);
+                log::debug!("[bn256 input]: {:?}", &input[64..96]);
+                log::debug!("[bn256 input]: {:?}", &input[96..128]);
```
Flouse: The `input_size` can be less than 128. This debug log may cause panic.

Flouse (changes requested):

crates/generator/src/syscalls/bn.rs
Flouse: Yes, that is intended.

As my understanding, when `px == Fq::zero() && py == Fq::zero()`,
it is a special coordinate at infinity with `z == P::Base::zero()`. 

similar implementation: https://github.com/openethereum/parity-ethereum/blob/c7f608ec740882eac94038249037ddd955b60d31/ethcore/src/builtin.rs#L408-L414

Flouse (dismissed):

Flouse (approved):

jjyr (approved):

PR #1033, opened 2023-03-27
feat: rpc-client retry on network error

blckngm:


crates/rpc-client/src/utils.rs
```diff
@@ -40,9 +41,94 @@ impl TracingHttpClient {
...
+                        continue;
+                    }
+                    return Err(e.context(format!("rpc {method}")));
+                }
+            }
+        }
+    }
+}
+
+pub struct ExponentialBackoff {
```
blckngm, line 69: Had to move this to avoid a circular dependency between `gw-utils` and `gw-rpc-client`.

jjyr (approved):

Simon-Tl (approved):

PR #1060, opened 2023-04-25
feat: timeout for rpc server

### LINKED ISSUE #1054: [BUG] currently does not support client body read timeout
From the code:

https://github.com/godwokenrises/godwoken/blob/2d8ad3510fcd528dfbfe76242b9572acdcf25084/crates/rpc-server/src/server.rs#L51

Seems `godwoken` also has a similar issue with `ckb`, https://github.com/nervosnetwork/ckb/issues/3880

I didn't verify it on `godwoken`, but from the code the `rpc` does not has a limit for a request, you can verify it with `curl`:

```console
curl --location 'http://127.0.0.1:3000/rpc' \
-H 'content-length: 100' \
--header 'Content-Type: application/json' \
--data-raw '   {
        "id": 42,
        "jsonrpc": "2.0",
        "method": "@ping",
        "params": [
        ]
   }'

```

An possible fix is to add a `timeout` middleware: https://github.com/chenyukang/jsonrpc-utils/blob/810f6a36eec3ce5be8673c11311bf1dcc506c3f9/examples/server-timeout.rs#L39

blckngm:
Body size limit is already provided by axum. Close #1054.


gwos-evm/polyjuice-tests/Cargo.toml
```diff
@@ -19,9 +19,6 @@ gw-traits    = { path = "../../crates/traits/" }
 gw-generator = { path = "../../crates/generator/" }
 gw-builtin-binaries = { path = "../../crates/builtin-binaries" }
 
-jsonrpc-utils = { version = "=0.2.0-preview.4", features = ["client"] }
-jsonrpc-utils-macros = "=0.2.0-preview.4"
-
```
blckngm, line 24: No longer necessary because any 0.2 version should be compatible.

Or we should use a lockfile for ployjuice-tests.

crates/rpc-server/Cargo.toml
```diff
@@ -44,6 +44,7 @@ jemalloc-ctl = { package = "tikv-jemalloc-ctl", version = "0.4.2" }
 jemalloc-sys = { package = "tikv-jemalloc-sys", version = "0.4.2" }
 errno = "*"
 tracing = { version = "0.1", features = ["attributes"] }
-jsonrpc-utils = "0.2.0-preview.2"
+jsonrpc-utils = { version = "0.2.0", features = ["server", "macros", "axum"] }
```
blckngm, line 47: Updated this while we are at it.

jjyr (approved):

zeroqn (approved):

PR #1070, opened 2023-05-16
fix: Using redis lua script to check & update rate limit key in an atomic operation

classicalliu:
Using lua script to avoid race condition between `expire` and `incr`

### DIFF
--- a/web3/packages/api-server/src/cache/guard.ts
+++ b/web3/packages/api-server/src/cache/guard.ts
@@ -13,6 +13,9 @@ export const EXPIRED_TIME_MILSECS = 1 * 60 * 1000; // milsec, default 1 minutes
 export const MAX_REQUEST_COUNT = 30;
 export const BATCH_LIMIT = 100000; // 100_000 RPCs in single batch req
 
+const RATE_LIMIT_SCRIPT =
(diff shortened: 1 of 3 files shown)


web3/packages/api-server/src/cache/guard.ts
```diff
@@ -55,79 +60,75 @@ export class AccessGuard {
...
+    // No limit for this RPC
+    if (maxNumber == null) {
+      return {
+        isOverRate: false,
+      };
     }
-    return false;
+
+    const result: [string, string] = (await this.store.evalsha(
+      await this.getEvalSha(),
```
blckngm, line 105: The script may become unavailable on redis due to e.g. `SCRIPT FLUSH`. You would need to reload it then. https://redis.io/docs/manual/programmability/eval-intro/#cache-volatility

classicalliu: image

Worker process will exit and restart when cache invalidation, this request will failed and will recover soon.

Cache invalidation is a small probability event, so we may not have to deal with it specifically ?

classicalliu: image

On second thought it would be better to handle the exception.

blckngm (approved):

jjyr (approved):

